PR #854, opened 2012-12-26
RFC: Custom block plugin system

tswsl1989:
This is just a proof of concept to try and get the ball rolling. The render definition for wool has been moved out of textures.py into a seperate file, which you can then add back in the config file using customblocks = ['wool']. This tries to render red wool using a custom texture, which I placed in /tmp/wool.png - you can download the file I used from http://sucs.org/~tswsl1989/texture-test.png

If you try this, then the texture will be visible, but very dark and hard to distinguish. I'm not sure if this darkening is the issue with the other wool colours.

Suggestions/comments?

### DIFF
--- a/customblocks/wool.py
+++ b/customblocks/wool.py
@@ -1,0 +1,41 @@
+#from overviewer_core import textures
+
+@material(blockid=35, data=range(16), solid=True)
+def wool(self, blockid, data):
+    if data == 0: # white
+        texture = self.terrain_images[64]
+    elif data == 1: # orange
+        texture = self.terrain_images[210]
+    elif data == 2: # magenta
+        texture = self.terrain_images[194]
+    elif data == 3: # light blue
+        texture = self.terrain_images[178]
+    elif data == 4: # yellow
+        texture = self.terrain_images[162]
+    elif data == 5: # light green
+        texture = self.terrain_images[146]
+    elif data == 6: # pink
+        texture = self.terrain_images[130]
+    elif data == 7: # grey
+        texture = self.terrain_images[114]
+    elif data == 8: # light grey
+        texture = self.terrain_images[225]
+    elif data == 9: # cyan
+        texture = self.terrain_images[209]
+    elif data == 10: # purple
+        texture = self.terrain_images[193]
+    elif data == 11: # blue
+        texture = self.terrain_images[177]
+    elif data == 12: # brown
+        texture = self.terrain_images[161]
+    elif data == 13: # dark green
+        texture = self.terrain_images[145]
+        texture = Image.new("RGBA", (16, 16), 0x99FFFFFF)
+    elif data == 14: # red
+        texture = self.terrain_images[129]
+        texture = Image.open("/tmp/wool.png").convert("RGBA").resize((16,16))
+    elif data == 15: # black
+        texture = self.terrain_images[113]
+    
+    return self.build_block(texture, texture)
+
--- a/overviewer.py
+++ b/overviewer.py
@@ -391,8 +391,10 @@ dir but you forgot to put quotes around the directory, since it contains spaces.
             worldcache[render['world']] = w
 
         # find or create the textures object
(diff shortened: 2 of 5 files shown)


Morketh:
Would it be possible to hook into Forge in someway (yes i realize this was written in Python and Forge in Java) i was looking at the way mods are loaded using Forge and it seems there is a Standard Block Registration API
" Blocks need to be registered to Forge. To do so, use GameRegistry.registerBlock(Block block, String internalName). Add this to the load method. " ~ [Forge Wiki](http://www.minecraftforge.net/wiki/Basic_Blocks#Block_Registration)
```java
GameRegistry.registerBlock(genericDirt, "genericDirt");
```
now im not really familiar with this sort of in-depth programming and im just kind of throwing ideas on the page but to me it seems like it should be easy (again this might be extremely hard) to define a way to "define" a block (and its textures) based on the way that Forge Loads those blocks (again falling back to the above post were i had mentioned they got loaded into the game and are visible on the client side). im really kind of stepping out of my normal field here to see what i can learn about this. Now correct me if im straying a bit far into the brush here but wouldnt a [bridge](https://github.com/CellProfiler/python-javabridge) help in being able to load these pre-existing APIs that come with Forge? seems if you were able to hook into that mod loader and ask for texture paths (based on the mods location) and then figure out what a block was (based on the way forge constructs the blocks) it would be a giant leap in the direction of mod support i see a lot of talk about waiting for an API, why not use the ones thats here?

CounterPillow:
Hi. Yes, block seems to have a standardised way of defining blocks, however, there are a couple of problems here. First off, the current block drawing code uses hand-written routines for constructing blocks for each id, with only some generalisation going on. This means that you can't throw arbitrary data at it and have it draw that. However, OIL will solve this™.

In the meantime, the official Mojang block model definitions have come out in 1.8 in some parts, but we still don't use it in OIL, mainly because nobody put in the effort so far to make it use the block model stuff. Nobody hasn't touched the code in a while, mainly because they didn't feel any interest in writing that code. Most Overviewer contributors don't play Minecraft anymore.

And herein also lies the problem with the suggested Forge support: None of us have ever used Forge, and likely won't in the near future (though I can't speak for anyone else). Minecraft modding has taken a bit of a tumble with 1.8 and some stuff surrounding that release, and it all seems very up in the air at the moment.

From the technical side, just quickly glancing over the bridge stuff I don't think that's the right approach, since you wouldn't want Overviewer to start your Minecraft server. I think the more appropriate solution would be writing a Forge plugin that exports the block data Forge handles in some way for us to use, but all that requires someone to do that work, and the master branch is too unsuited for that sort of stuff while the oil (or is it oil-py3k?) branch is too far from complete to reduce the amount of outstanding work significantly.

Sorry :(

Morketh:
Well that line of reasoning really gave me something to think about. Whats the status of Oil? Percent done, todo list, whats been done so far? i like your idea of having a forge side to Overviewer ill have to roll that idea around a bit and see what i can come up with. that seems like a more a modular solution seeing as forge keeps a dictionary of different blocks,items,liquids.what section of overviewer actualy handles the blocks? (the hand written parts you mentioned) maybe that will help me see whats going on in engine so i can toss a few wrenches at it

CounterPillow:
The imaging code and rendering code of OIL is "done" as far as I know. It does not read block models from Minecraft as far as I'm aware, and a lot of the special case handling for the data values is not implemented. Furthermore, it's still missing other parts of a general rewrite we wanted to do, such as switching away from Google maps, a better POI system, a new config file format and so on.

The currently handwritten block definition parts of Overviewer are in the infamous `textures.py` file in the `overviewer_core` directory. Furthermore, some special case handling in terms of lighting is done in the render primitives C code which can be found in the `src` directory.

3nt3:
did anything happen to this issue?

tswsl1989:
> did anything happen to this issue?

Not really. There were some odd issues with this proof of concept and since then the project has seen a lot of improvements.
The basic approach should still be viable if anyone wants to pick it up, but it's not likely to be something I do at this point I'm afraid.

PR #1121, opened 2014-06-24
Add new 1.7.2 biomes and update colors

### LINKED ISSUE #1104: Biomes list is incomplete
It is said Overviewer supports Minecraft v1.7.

However, biomes list as specified in https://github.com/overviewer/Minecraft-Overviewer/blob/master/overviewer_core/src/primitives/biomes.h is incomplete
For example, 'Flower Forest' is non-existent.

A biome list might be found here: http://minecraft.gamepedia.com/Biome

dbergl:
I think this resolves #1104.

I added these way back when 1.7.2 came out but forgot about it. I think a couple of the existing biome colors likely changed when i updated the temp/humidity so this might mean a full new render would need to happen.


dbergl:
I have reverted the changes that affected the biome colors and fixed a couple biome names

agrif:
June is the season of productivity!

We've merged these changes, and we'll keep an eye on a nice time to introduce the more breaking but correct biome changes. Thanks!

Issue #787, opened 2012-08-16
MineralOverlay cannot specify alpha


HTL2001:
Can't control alpha channel on MineralOverlay.  I suggest something like depth tinting or allow a constant to be applied.

Issue #556, opened 2011-12-06
"Railway overlay"


danishduckling:
Hi, would it be possible, in the future to make an overlay of all kinds of rails (powered, detector and regular), that are ontop of a specific block? (say, you have a standard on your server that says, all railways are placed on gravel), would it be possible to make a overlay, of rails that are placed on the specified block? (not rendering the block but only the rail ontop of it)

PR #1152, opened 2014-09-03
Adding StructureOverlay an overlay to color the map according to structures

enaut:
Differences to my earlier Mineraloverlay extension pull request:
  * While writing documentation it occured to me that the extended MineralOverlay was too complex
  * I split of from MineralOverlay the StructureOverlay
  * I moved the default input from C where it was quite awful to python where it is much nicer

So actually this closes the former pull request #1125.

As for the code here it is:

A usecase to demonstrate a possible application of the extended functionality:

**"Rails Overlay that draws only the rails that are on Cobblestone for a subway map."**

With this patch it is very easy to achive that:

```python
MineralOverlay(minerals=[(((0, 0, 0, 66), (0, -1, 0, 4)), (255, 0, 0, 255)),
                         (((0, 0, 0, 27), (0, -1, 0, 4)), (0, 255, 0, 255))])
```

In this case the overlay will be red for rails on cobblestone and green for powerrails on cobblestone.
The syntax is `(<tuple of conditions>, <target color>)`
 * where `<target color>` is a 4 tuple with a `(r, g, b, a)` color
 * and `<tuple of conditions>` is a tuple with an arbitrary number of conditions with the following syntax:
`((relx, rely, relz, blkid), ...)` where the `rel<>` parameters specify the relative coordinates to the block that is checked if it matches bklid.

In the example the fist tuple `(0,0,0,66)` checks if at the current position is a
rail while `(0,-1,0,4)` checks if at one below the current position is a cobblestone.
If both are true then the color `(255, 0, 0, 255)` is used.

A Sample Config file exploiting the capabilities:

``` python
worlds['My World'] = "~/.minecraft/saves/test/"
outputdir = "/tmp/test_render"
rendermode = "lighting"

renders["render1"] = {
    'world': 'My World',
    'title': 'A regular render',
}
renders["render_overlay_dafault_rails"] = {
    'world': 'My World',
    'title': 'Default Rails',
    'rendermode': [ClearBase(), StructureOverlay()],
    'overlay': ['render1'],
}
renders["render_overlay_cust_rails"] = {
    'world': 'My World',
    'title': 'Custom Rails',
    #relative coordinates [[(relx, rely, relz, mineral)], (red, green, blue, alpha)]
    'rendermode': [ClearBase(), StructureOverlay(structures=[(((0, 0, 0, 66), (0, -1, 0, 4)), (255, 0, 0, 255)),
                                                            (((0, 0, 0, 27), (0, -1, 0, 4)), (0, 255, 0, 255))])],
    'overlay': ['render1'],
}
```

The "Default Rails" overlay uses default coloring of the structures overlay. "Custom Rails" uses some custom coloring.

fixes overviewer/Minecraft-Overviewer#556 and fixes overviewer/Minecraft-Overviewer#787

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1099,6 +1099,32 @@ MineralOverlay
 
             MineralOverlay(minerals=[(64,(255,255,0)), (13,(127,0,127))])
 
+StructureOverlay
+    Color the map according to patterns of blocks. With this rail overlays
+    or overlays for other small structures can be realized. It can also be
+    a MineralOverlay with alpha support.
+
+    This Overlay colors according to a patterns that are specified as
+    multiple tuples of the form ``(relx, rely, relz, blockid)``. So
+    by specifying ``(0, -1, 0, 4)`` the block below the current one has to
+    be a cobblestone.
+
+    One color is then specified as
+    ``((relblockid1, relblockid2, ...), (r, g, b, a))`` where the
+    ``relblockid*`` are relative coordinates and the blockid as specified
+    above. The ``relblockid*`` must match all at the same time for the
+    color to apply.
+
+    Example::
+
+        MineralOverlay(minerals=[(((0, 0, 0, 66), (0, -1, 0, 4)), (255, 0, 0, 255)),
+                                 (((0, 0, 0, 27), (0, -1, 0, 4)), (0, 255, 0, 255))])
+
+    In this example all rails(66) on top of cobblestone are rendered in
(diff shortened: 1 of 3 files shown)


agrif:
Awesome, awesome. Thank you for this! Sorry for the long wait.

enaut:
Thank you for accepting and good to know that this work is appreciated!
As for the long wait, it certainly would help to get some indicator that someone cares even if it is not integrated right away but I can live with just waiting too ;)...

Issue #1175, opened 2014-10-28
regions.js not working ? 


T-G-R:
Why in the web-directory containing the file regions.js? the change does not result in any effect.

agrif:
regions.js is (itself) now working. There's currently no way to use it with genPOI, but there are already other issues about that so I'm closing this.

Issue #889, opened 2013-02-25
Uncaught TypeError: Cannot call method 'push' of undefined worldguard.js.php:2 (anonymous function)


Jakky89:
I am using a slightly modified worldguard region markers php script to generate a javascript array of all regions from database. This is the part that doesn't seem to work as described in the overviewer addons directory:
```php
        while ($wg_cuboid_row=mysqli_fetch_row($wg_cuboid_res)) {
            $color = $color_normal;
            $label = $wg_cuboid_row[0];
            $minx = $wg_cuboid_row[2];
            $miny = $wg_cuboid_row[3];
            $minz = $wg_cuboid_row[4];
            $maxx = $wg_cuboid_row[5];
            $maxy = $wg_cuboid_row[6];
            $maxz = $wg_cuboid_row[7];
            $cube = false;
            if(abs($miny-$minx)>3) {
                $cube = true;
            }
            //center
            $output .= '{"label": "' . $label . '-top", "color": "' . $color . '", "opacity": 0.5, "closed": true, "path": [';
            $output .= '{"x": ' . $minx . ', "y": ' . $maxy . ', "z": ' . $minz . '},';
            $output .= '{"x": ' . $minx . ', "y": ' . $maxy . ', "z": ' . $maxz . '},';
            $output .= '{"x": ' . $maxx . ', "y": ' . $maxy . ', "z": ' . $maxz . '},';
            $output .= '{"x": ' . $maxx . ', "y": ' . $maxy . ', "z": ' . $minz . '}';
            $output .= ']},';
            if($cube) {
                //isometric cube draw
                $output .= '{"label": "' . $label . '-isoleft", "color": "' . $color . '", "opacity": 0.5, "closed": true, "path": [';
                $output .= '{"x": ' . $minx . ', "y": ' . $miny . ', "z": ' . $minz . '},';
                $output .= '{"x": ' . $minx . ', "y": ' . $miny . ', "z": ' . $maxz . '},';
                $output .= '{"x": ' . $minx . ', "y": ' . $maxy . ', "z": ' . $maxz . '},';
                $output .= '{"x": ' . $minx . ', "y": ' . $maxy . ', "z": ' . $minz . '}';
                $output .= ']},';
                $output .= '{"label": "' . $label . '-isoright", "color": "' . $color . '", "opacity": 0.5, "closed": true, "path": [';
[...]

Issue #642, opened 2012-03-10
Will regions be added in the future


redorkulated:
Hey, I know that the current version is beta and lots of stuff is not implemented because of that. I am just wondering will regions be added back onto the maps again, those created by regions.js? On the blog it mentions "Signs, Spawn markers, and other Points of Interest" and (its probably me being dumb) but do the regions count as Other Points of interest :D. If they are being added will they function in the same way where lines or shapes can be drawn? If so could it possible to define line thickness in the region,js per region?

agrif:
Yet one year later-er, and `regions.js` is now being loaded and works with the toggles. There's no interface for generating regions a la genPOI, but it's a start.

Shuttle099:
Any updates?

CounterPillow:
@Shuttle099 it's possible to display regions by adding them to the regions.js, but it's not documented. So no, no updates so far.

Shuttle099:
I found the region.js but how do I go about using it?

simon511000:
otherwise we can create regions using the Leafletjs documentation:
![image](https://user-images.githubusercontent.com/31013205/90722089-6b499400-e2ba-11ea-8bae-b28d6830aa22.png)
![image](https://user-images.githubusercontent.com/31013205/90722123-76042900-e2ba-11ea-8765-309f5edec0a7.png)

PR #1186, opened 2014-12-13
regions.js fixed

kiskoza:
Fixed regions.js, added an example. Now you can draw polylines and polygons on the map. Works for #1175 and #642 . Maybe helps with #889


agrif:
Thank you for this! I have no idea how we let this sit so long.

DerTod2:
Now we needs only the possibility to load the regions out of an json and auto-refreshes this so plugins could update the regions json and the overviewer automatically shows the changes <3

Currently we uses dynmap to show regions and players in realtime... but MinecraftOverviewer is so much better than dynmap :/

PR #1207, opened 2015-02-11
Docs import change

tdierks:
This code is correct, but doesn't work correctly with signs since they got JSON text formats in 14w25a. I will try to create another PR with an improved sign function tomorrow, but it's turning out to be complex, so for the moment, I'll stick with this tip on making imports work.


agrif:
Great, thank you!

PR #1208, opened 2015-02-15
OrderedDict removal && setup.py cleanup 

gmr:
Update ``setup.py`` to use ``setuptools.setup`` and refactor a bit, including pep8 cleanup and dependency installation. ``distutils`` use is now deprecated in favor of ``setuptools`` (again), though I did not refactor it out completely. A lot of the refactoring is done in the hope that I can get it to building wheel files which will allow platform specific binary distribution without the build step via pip.

Use the 3rd party odict module available on pypi instead of having the code in util.py. When ``setup.py develop`` or ``setup.py install`` are invoked for Python 2.6, odict will be installed, so the fall through in failures should not happen. The only reason we pass on AttributeError is because to invoke ``setup.py``, ``overviewer_core.util`` needs to be imported.

Update version checking in overviewer.py to match how setup.py has been updated.

This also updates ``.travis.yml`` to speed up builds adding a requirements.txt and requirements-2.6.txt file that can be used with travis-ci for caching purposes. It also enabled containerized builds in travis. Additionally remove commented out things for 3.2. They're in git history if they need to be added back/uncommented.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - wget http://hg.effbot.org/pil-117/raw/f356a1f64271/libImaging/ImPlatform.h
 install:
   - pip install -q pillow
-  - pip install -q numpy
+  - python setup.py develop
   - python setup.py build
 before_script:
   - git clone git://github.com/overviewer/Minecraft-Overviewer-Addons.git ~/mcoa/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,16 @@
+sudo: false
 language: python
 python:
   - "2.6"
   - "2.7"
-  # - "3.2"
 env:
   - MC_VERSION=1.8
-before_install:
-  - wget http://hg.effbot.org/pil-117/raw/f356a1f64271/libImaging/Imaging.h
-  - wget http://hg.effbot.org/pil-117/raw/f356a1f64271/libImaging/ImPlatform.h
 install:
-  - pip install -q pillow
-  - python setup.py develop
+  - mkdir -p Imaging/libImaging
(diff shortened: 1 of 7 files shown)


eminence:
Cool!  Did you happen to do any testing with Windows and/or py2exe?

gmr:
I did not, I'll startup a VM... I didn't really change any of those bits.

eminence:
if it's easy for you, go for it.  if not, don't worry about it, i'll get the build-bots to do a little testing

CounterPillow:
I'm not so sure about the ordered dict one. Is there a specific reason for relying on an external dependency when we already have an implementation in Overviewer? If this makes packaging and installation on systems such as CentOS5/6 (which AFAIR are the only ones still using python 2.6) harder, then I'm not really in favour of this change.

As for Windows, I remember that distutils was various shades of broken for MinGW, especially MinGW64, so factoring it out is a move I approve of. However, only as long as any new systems are either less or equally broken in this regard. Since MSVC is what we use for the official Windows builds, MSVC probably has priority though.

agrif:
We try to avoid external dependencies, mostly because adding them requires them to have packages easily available on debian and redhat systems, which pretty much restricts us to the most popular ones. In particular, it appears that odict does not have a debian package. Wheel files would be interesting, but I don't think they'd replace the native package repositories.

A *lot* of this is stuff I wanted to do on the oil branch, but there's still a lot of fundamental work to be finished there before it starts making sense to clean up the packaging.

gmr:
I don't have time to update this, so I'll close it and let it bit-rot instead of litter your issue/pr list :)

Issue #1205, opened 2015-02-10
Overviewer renders ice bright in night view (not smooth_night!)


shm512:
Hi there,
There was an issue #1032 about ice being rendered with a bright color (~ #6584D6) in night view. It has been solved, but only for smooth_lighting, while with default lighting it's still too bright:
http://fgsfds.dqteam.org:8000/
I am using version built on Sat Feb  7 22:01:13 2015
on Arch Linux x86_64.

agrif:
Wow, this was an amazingly old bug. Should be fixed as of 0422b0e711b3d57f1825c8d7746cccb3df0d7f0a. Thanks to CounterPillow for poking me on this.

CounterPillow:
A fix for this has been committed by agrif. To force a rerender of your entire map to see the changes to old tiles, you can use the `--forcerender` option, or put `"forcerender":True` in the render dict entries for the renders you want forcefully updated, but be sure to remove it again afterwards.

Nighthawk200:
Apologies for reviving a closed topic, but how do you generate lines between markers like that in your rendered map, shm512? I've been wanting to use such a feature in my own map for a while, but can't seem to find any mention of how to do it anywhere in either the documentation or here on Github...

CounterPillow:
That's the polyline feature, it's currently still not documented but I think the PR that added it contained an example configuration.

Nighthawk200:
Thanks for the quick reply. I found the PR with the example (https://github.com/overviewer/Minecraft-Overviewer/pull/883) ... unfortunately the example is a bit counter-intuitive for me to follow. The example there appears to be generating a polyline between two detected POIs (command blocks with a /tp command); whereas my idea is to manually define multiple different-coloured polylines that show up with a set of manual POIs defined in my config file (railway lines for railway stations - similar to what the original poster of this topic has done).

deltaidea:
@Nighthawk200, author of the config for that server here. We host it here: [deltaidea/overviewer-config](https://github.com/deltaidea/overviewer-config)
Here's how we've done it:

- find out that there's an undocumented `"polyline"` option in PoIs that looks like this:

  ```Python
  {
    ...
    "polyline": tuple(
      { "x": ..., "y": ..., "z": ... },
      { "x": ..., "y": ..., "z": ... },
      ...
    )
  }
  ```
[...]

PR #1211, opened 2015-02-23
fix lighting for flowing water and ice (issue #1205)

agrif:
fix for issue #1205


CounterPillow:
Tested and confirmed for working.

Issue #1198, opened 2015-01-17
Newly placed Signs not working correctly in POI.


FidgetyRat:
My server noticed lately that new signs aren't being caught by our filtering rules. I saw in another thread to change from poi['Text1'] == "Location" to "Location" in poi['Text1']. Now the signs do show up, but they contain a lot of extra tagging and information!

>def locationFilter(poi):
    from fnr_additions import correctMinecraftText
    if poi['id'] == 'Sign' and "location" in correctMinecraftText(poi['Text1']).lower():
        return "\n".join([correctMinecraftText(poi['Text2']),
                          correctMinecraftText(poi['Text3']),
                          correctMinecraftText(poi['Text4'])])

Result:
>{extra:[Temporary],text:}
  {extra:[Ender Portal],text:}
  {extra:[Entrance],text:}

Note: The import just strips out some extra characters like quotations and such and works fine on our old signs.

Old signs parse just fine. I believe the "Old" signs were added during 1.8 with overviewer from 11/2014 and the new signs in 1.8.1 with overviewer from 01/08/2014. The only other changes are that we are also using Spigot on the server, but I can't comment as to whether spigot is adding the "Extra" stuff or not.

For now I added the "{extra:[" and "],text:}" to my library, so they get stripped out, but figured maybe this will hit someone else as well. For the longest time I couldn't figure out why my signs were not showing up at all.

CounterPillow:
Seems like a duplicate of #1188

Lomsor:
This isn't a duplicate of #1188 
I have the same issue. It stems from Spigot, with Vanilla this doesn't occur.
I looked into the NBT and the text is stored like this:
{"extra":["YOUR TEXT"],"text":""}

This only occurs on new signed placed while running Spigot 1.8

CounterPillow:
You're right in that it's not an exact duplicate; spigot goes a step further to break things and puts the text in the "extra" field instead of the parent "text" field of the JSON, but in the end it still depends on #1188.

Properly solving #1188 will also solve this issue here though, since `"YOUR TEXT"` is considered a "valid" JSON object in this format.

So for all intents and purposes, it is a duplicate, as it properly conforms to the minecraft JSON text standard.

j3rd:
A workaround using reg-ex, tested in python app not overviewer.

```python
import re

def signFilter(poi):
    if poi["id"] == "Sign":
[...]

PR #1215, opened 2015-03-07
[genPOI] Work around JSON signs

### LINKED ISSUE #1188: unicode character issues from minecraft data (since 1.8?)
I'm using a modified version of the script in the documentation for showing signs as POIs on the map:

    def signFilter(poi):
        if poi['id'] == 'Sign':
            return "\n".join([poi['Text1'], poi['Text2'], poi['Text3'], poi['Text4']])

It appears that many characters are being encoded as unicode strings (`\u0027` instead of an apostrophe for example). It sounds like this is going on in various places as I've seen tickets related to this elsewhere due to the minecraft data files now being in UTF8 format rather than ASCII by default now.

CounterPillow:
Fixes #1188 and #1198 (which is a duplicate of the former but you guys haven't closed it because you're lazy bums)

### DIFF
--- a/overviewer_core/aux_files/genPOI.py
+++ b/overviewer_core/aux_files/genPOI.py
@@ -24,6 +24,7 @@ import urllib2
 import multiprocessing
 import itertools
 import gzip
+import json
 
 from collections import defaultdict
 from multiprocessing import Pool
@@ -44,6 +45,35 @@ def replaceBads(s):
         x = x.replace(bad,"_")
     return x
 
+# If you want to keep your stomach contents do not, under any circumstance,
+# read the body of the following function. You have been warned.
+def jsonText(s):
+    if (s.startswith('"') and s.endswith('"')) or \
+        (s.startswith('{') and s.endswith('}')):
+        try:
+            js = json.loads(s)
+        except ValueError:
+            return s
+
+        def parseLevel(foo):
+            bar = ""
+            if isinstance(foo, list):
+                for extra in foo:
+                    bar += parseLevel(extra)
+                return bar
+            if isinstance(foo, dict):
+                if "text" in foo:
+                    bar += foo["text"]
+                if "extra" in foo:
+                    bar += parseLevel(foo["extra"])
+                return bar
+            elif isinstance(foo, basestring):
+                return foo
+
+        return parseLevel(js)
+
+    else:
+        return s
 
 # yes there's a double parenthesis here
 # see below for when this is called, and why we do this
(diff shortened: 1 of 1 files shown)


CounterPillow:
Apparently this PR is broken because I treat the sign text contents as JSON, when it's actually yaml or some shit like that! Hooray! Of course, everyone in the Minecraft community is too braindead to know the difference between `{"text":"foo"}` and `{text:"foo"}` so they call it "JSON" anyway. Will fix sometime in the future. Maybe.

CounterPillow:
As I've just discovered it's neither YAML nor JSON, it's their own thing apparently, which is so close to JSON you mistake it for JSON if you don't look closely, but it violates the JSON spec just for the fun of it and thus you can't use a JSON parser to parse it. Because why use existing general purpose formats when you can make up your own, even if you've already made up your own! Just make up another one and then store it inside the one everybody already speaks, so they have to go through extra hoops and write a second parser. For literally no reason.

Sorry for being so rant-y, but I have no formal education in the field and have never professionally written code before, but even I can see from a mile away that this is the most ass-backward shit ever.

CounterPillow:
Okay, interesting discovery just now.

Apparently, while what you type into Minecraft to generate it is not JSON, Minecraft seems to actually store it as valid JSON! So my PR is correct?

Here's some libredstone nbttool output:
```
TAG_List("TileEntities"): 1 entries of type TAG_Compound
{
    TAG_Compound: 8 entries
    {
        TAG_Int("x"): 2
        TAG_Int("y"): 104
        TAG_Int("z"): 216
        TAG_String("Text4"): null
        TAG_String("Text3"): null
        TAG_String("Text2"): null
        TAG_String("Text1"): {"extra":["there"],"text":"Hello"}
        TAG_String("id"): Sign
    }
}
```

That is JSON, with the double quotation marks and all.

And again, just to confirm I'm not dreaming, in colour blue:

```
TAG_String("Text1"): {"color":"blue","text":"Hello there"}
```

Yeah so this PR is correct after all, Mojang's non-existent documentation is just not accurate because it doesn't exist.

Feel free to merge.

CounterPillow:
Just fixed another corner case which would spam null's all over our precious signs. The reason for that is that since we do not know whether a sign is JSON or not, "null" might be both the text "null" on a 1.7 sign or just `null` as a JSON value for an empty line. We're now doing what Minecraft does, which is eating all nulls. Refer to the comment in the code if you feel like headdesking at how retarded this shit is. Yes, they're even doing it for lists. No, we don't reproduce that behaviour because it needlessly breaks too much shit from previous versions while never really occurring in JSON signs.

agrif:
*eyes gloss over wall of text*

PR #1247, opened 2015-08-17
add support for advpng to optimizeimages.py

mc680x0:
Signed-off-by: Horst Burkhardt <mc@680x0.com>

Use of advpng in conjunction with optipng can frequently provide better compression than either tool alone, this commit permits use of advpng by the overviewer.


mc680x0:
please nuke this, i've fixed it since

PR #1248, opened 2015-08-17
add support for advpng to optimizeimages.py

mc680x0:
AdvPNG, working in conjunction with OptiPNG, can produce better results than either tool alone; this commit permits advpng to be used from the overviewer.

### DIFF
--- a/overviewer_core/optimizeimages.py
+++ b/overviewer_core/optimizeimages.py
@@ -121,6 +121,19 @@ class optipng(Optimizer, PNGOptimizer):
     def is_crusher(self):
         return True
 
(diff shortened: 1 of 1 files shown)


eminence:
FYI:  you didn't need to close your first PR (#1247) -- you can just push more commits to the branch and the PR will update automatically.

mc680x0:
Ah, I didn't realise github would presume that. Thanks.

awilfox:
:+1: Every little bit helps on data sets this large, it works here.

agrif:
Half a year later, we've merged this code. Sorry for the wait!

Thank you for your contribution!

PR #1254, opened 2015-10-18
Add region directory to corruption warnings

CounterPillow:
I'm not sure if this is the correct approach to solve this, since the code appears to be all over the place. Should this be rewritten to have one central "hey, something is wrong with this world" handler which logs the warning with better context?


agrif:
It probably should be central, but this will do for now. Thanks!

Issue #1262, opened 2015-12-05
The new Path block is not being renderd


MakerTim:
We got a Overviewer at our site, and the special-render-blocks like Slimeblock and Path block is not being renderd at all. 
I was watching the config and some requests for it, can there be an option that this kind of blocks just render as full blocks, i dont care that there not transparant or 1/16 short, but this doesnt look at all what it supposted to look like!
http://render.imine.nl/iMineS3/#/-2026/64/1075/max/0/0

CounterPillow:
Notes for anyone wanting to implement this:

* ID: 208
* Like full block but only 15/16th in height.
* Added in 15w31 (to be version 1.9), probably will be in stable in January.

CounterPillow:
fixed in 5f9276a24a8d64efe4b91cf29c565b4541a0cdf0

PR #1263, opened 2015-12-06
#1262-fix

robbix1206:
Add of the grass path block and correct the farmland to a 15/16 block.
I think its correct but well.. it's my first commit so please if someone can verify what I done.


overviewer_core/textures.py
```diff
@@ -2167,13 +2167,41 @@ def crops(self, blockid, data):
...
-@material(blockid=60, data=range(9), solid=True)
-def farmland(self, blockid, data):
-    top = self.load_image_texture("assets/minecraft/textures/blocks/farmland_wet.png")
-    if data == 0:
-        top = self.load_image_texture("assets/minecraft/textures/blocks/farmland_dry.png")
-    return self.build_block(top, self.load_image_texture("assets/minecraft/textures/blocks/dirt.png"))
+# farmland and grass path (15/16 blocks)
+@material(blockid=[60,208], data=range(9), solid=True)
+def farmland_and_grass_path(self, blockid, data):
+    if blockid=60:
```
agrif: You probably mean `if blockid == 60` here. Other than that, I can't see anything obviously out of place.

robbix1206: Thanks you haven't seen it ><
I took the daylight_sensor as model and copy it and modify it.

robbix1206:
Doesn't interference why texture findre crash ><

robbix1206:
Was thinking of Travis configure for 1.8 at first but after looking at the log overviewer can't access to the grass picture ><

agrif:
Since this is a 1.9 block, we'll either merge this in a snapshot branch to eventually be merged when 1.9 is released, or just wait to merge this when 1.9 is released.

Either way, travis should work itself out once it actually uses the 1.9 jars.

NoxNebula:
The path Blocks are not 15/16, it seems they are 1/2 block, as high as a slab.

eminence:
I'm looking at http://minecraft.gamepedia.com/Grass_Path which says they are 15/16th of a block.  Has that changed?

NoxNebula:
I mean, they are 1/2 at the overviewer, but in minecraft 15/16

eminence:
Closed via 5f9276a24a8d64efe4b91cf29c565b4541a0cdf0

NoxNebula:
They are now ok :+1:

Issue #1266, opened 2015-12-27
corrupt uuidcache causes genpoi to crash


eminence:
- [x] when writing the uuidcache, write to a tmp file then move it into place
- [x] when loading the cache, catch both gzip and json decode errors

PR #1267, opened 2015-12-27
genpoi UUID improvements

eminence:
* When reading the cache, catch some errors on load, instead of crashing
* When writing to cache, write to tmp file, then move it into place.
  This should be more robust if a ctrl+c is recieved while writing the
  cache

Addresses #1266 

cc @ion1 @agrif 

I'll merge this in a few days, so let me know if you have any comments

### DIFF
--- a/overviewer_core/aux_files/genPOI.py
+++ b/overviewer_core/aux_files/genPOI.py
@@ -33,6 +33,7 @@ from optparse import OptionParser
 from overviewer_core import logger
 from overviewer_core import nbt
 from overviewer_core import configParser, world
+from overviewer_core.files import FileReplacer
 
 UUID_LOOKUP_URL = 'https://sessionserver.mojang.com/session/minecraft/profile/'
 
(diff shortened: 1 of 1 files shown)


overviewer_core/aux_files/genPOI.py
```diff
@@ -220,9 +233,13 @@ def load_cache(cls, outputdir):
     @classmethod
     def save_cache(cls, outputdir):
         cache_file = os.path.join(outputdir, "uuidcache.dat")
-        gz = gzip.GzipFile(cache_file, "wb")
-        json.dump(cls.uuid_cache, gz)
-        logging.info("Wrote UUID cache with %d entries", len(cls.uuid_cache.keys()))
+        try:
+            gz = gzip.GzipFile(cache_file + ".tmp", "wb")
+            json.dump(cls.uuid_cache, gz)
+            os.rename(cache_file + ".tmp", cache_file)
```
CounterPillow: Shouldn't we use stuff from files.py in here in case rename doesn't work?

agrif: I concur! There's a whole thing for this IIRC.

agrif:
About saving the corrupted uuid cache: do we ever actually expect somebody to ungzip the file and fix it? Or fix it so it's un-gzippable? Why not just delete the cache?

eminence:
Someone from IRC convinced me that silently deleting files wasn't a good idea.  It was a weak argument, and the only plausible explanation I saw was: someone accidentally copied important data into the UUID cache.

overviewer_core/aux_files/genPOI.py
eminence: Is this the library you are referring too?  How do you all feel about adding a new 3rd party dependency?

https://pythonhosted.org/files/

CounterPillow: no, I'm talking about files.py in overviewer_core... which, uh... I think you wrote.

eminence: hey neat! i forgot about that!  ok yeah, we should use that

PR #1276, opened 2016-02-13
Remove inline for estimate_blocklevel

Mirclus:
Inlining the function allows the compiler to optimize away the
function completely. Clang 3.7.1 does exactly that.

This leads to an error, if the library is used with python:

    % ./overviewer.py
    Traceback (most recent call last):
      File "/tmp/minecraft/Minecraft-Overviewer/overviewer_core/__init__.py", line 20, in check_c_overviewer
        import c_overviewer
    ImportError: /tmp/minecraft/Minecraft-Overviewer/overviewer_core/c_overviewer.so: undefined symbol: estimate_blocklevel

    Something has gone wrong importing the c_overviewer extension.  Please
    make sure it is up-to-date (clean and rebuild)

### DIFF
--- a/overviewer_core/src/primitives/lighting.c
+++ b/overviewer_core/src/primitives/lighting.c
@@ -95,7 +95,7 @@ calculate_light_color_fancy_night(void *data,
  * may (and probably should) pass NULL.
  */
 
-inline unsigned char
+unsigned char
 estimate_blocklevel(RenderPrimitiveLighting *self, RenderState *state,
                          int x, int y, int z, int *authoratative) {
 


CounterPillow:
We probably use the inline keyword way too liberally anyway, the compiler should usually be smart enough to know when to make the call.

Interesting though that Overviewer compiles with clang. You're probably the first person to do so.

Thanks for the PR, I will test it later.

CounterPillow:
Interestingly, `estimate_blocklevel` is only ever called in lighting.c, so I'm not even sure why anything would look for that symbol if it was completely inlined away. Not sure what's going on there. Maybe @agrif can explain.

agrif:
Overviewer compiles on OSX, which has been using clang for a while. I've never seen this before. I have no idea what's causing it.

Even with the inline keyword, the compiler shouldn't be removing the symbol if something refers to it by name. Does this still happen with a completely clean build?

Mirclus:
Yes, it is still reproducible and I think I know whats happening, as I did some digging:

Both GCC (v5.3.0) and Clang (v3.7.1) emit [a warning](https://gist.github.com/Mirclus/1ebed3c251e6b9e3dabb) on compiling _lightning.c_: `lighting.c:138:18: warning: ‘get_data’ is static but used in inline function ‘estimate_blocklevel’ which is not static`

The C standard distinguishes 2 types of functions: internal linkage (`static`) and external linkage (`extern` keyword, optional). So, a function without an explicit mention, a function has external linkage, therefore `estimate_blocklevel` has external linkage. If a function with external linkage contains an `inline` specifier, no code is generated. It is just a hint for the compiler to use this implementation, if he chooses to inline the function. There needs to be a non-inline definition somewhere.

In this case, gcc (and clang on your system) chose to inline the function and then doesn't need the definition anymore. Clang on my system (Arch Linux) does not inline the function. Therefore the library misses a definition and on load of the library this leads to an unresolved external.

So, another fix would have been to add a static modifier to the function and keep the inline.

Issue #1144, opened 2014-08-25
northdirection + crop results in wrong coordinates and compass in output


CounterPillow:
As reported by two people on IRC, cropping and changing the northdirection results in wrong (flipped) coordinates for the world, and also a wrong compass.

Config one:

```python
worlds["world"] = "/home/DOKDOR/1.7.9/world"
 
renders["normalrender"] = {
    "world": "world",
    "title": "PvE daytime",
    "rendermode": "normal",
    "dimension": "overworld",
    "crop": (-2000, -2000, 2000, 2000),
    "northdirection":"upper-left",
    "imgformat":"jpeg",
    "imgquality": 70
}
 
renders["reversednormalrender"] = {
    "world": "world",
    "title": "PvE daytime Reversed",
    "rendermode": "normal",
    "dimension": "overworld",
    "crop": (-2000, -2000, 2000, 2000),
    "northdirection":"lower-right",
    "imgformat":"jpeg",
    "imgquality": 70
}
 
outputdir = "/var/www/html"
```

Config 2:

```python
from optimizeimages import optipng

worlds["world"] = "/srv/minecraft/world"
processes = 4
northdirection = "upper-right"

renders["Day"] = {
    "world": "world",
    "title": "Day",
    "rendermode": smooth_lighting,
    "dimension": "overworld",
    "crop": (-50000, -50000, 50000, 50000),
    "optimizeimg": [optipng()]
}

outputdir = "/my/obfuscated/outputdir/"
showspawn = False
```

Both cases were reported after the recent changes of the cropping option merged in, so perhaps it is related.

tdelcourt:
Ive noticed the same thing.

```python
renders["walie8-crop-smooth_lighting-ul"] = {
    "world": "walie8",
    "title": "walie8-smooth_lighting-ul",
    "rendermode": smooth_lighting,
    "dimension": "overworld",
    "northdirection": "upper-left",
    "crop": (0, 0, 100, 100),
}
renders["walie8-crop-smooth_lighting-ur"] = {
    "world": "walie8",
    "title": "walie8-smooth_lighting-ur",
    "rendermode": smooth_lighting,
    "dimension": "overworld",
    "northdirection": "upper-right",
    "crop": (0, 0, 100, 100),
}
```
Gives me two different spawn points (248, 211) and (261, 224) and two different and non-overlapping pieces of land (albeit verry close to each other)

[...]

PR #1282, opened 2016-03-05
Fix render when using rotation and crop

Socolin:
The rotation should be applied after the crop. This order was changed in this commit https://github.com/overviewer/Minecraft-Overviewer/commit/b6ac54a2b61d0b8f45adab36ac5179e83bcef710


Socolin:
It fix issue #1144

agrif:
Ah, that's a pretty subtle one. Thank you for your contribution, even if it took us a while to get to it!

Issue #1280, opened 2016-02-19
missing textures for 1.9 snapshots / prereleases


NoxNebula:
The overviewer is searching the textures following textures in 1.9 snapshots / prereleases:
- [x] 'assets/minecraft/textures/blocks/redstone_dust_line.png'
- [x] 'assets/minecraft/textures/blocks/redstone_dust_cross.png'
- [x] 'assets/minecraft/textures/blocks/command_block.png'

I have to manually copy these textures from 1.8.* to the snapshot texturepath, in order to make the overviewer work.

`redstone_dust_cross` has been removed,
`redstone_dust_line` has been renamed to *0 respectively *1
`command_block` has now different textures, the best replacements should be command_block_back and command_block_front.

- [x] pathblocks are also not rendered (i use the git version of overviewer).

please fix

xd1936:
Now that 1.9 is officially out, this should be very high priority.

eminence:
Please give the latest commit from the `master` branch a test!

seeseekey:
Hello,

i have tested the new version. The new end blocks renders well, but command blocks are not correctly rendered. There are empty on the map.

SteseDav:
I'm getting the same problem on a Windows XP machine (don't ask), with 1.8.9 and 1.9 versions of Minecraft installed by the launcher. If the above fix works, when would it be normally included in the pre-compiled windows download?

Yoplitein:
Grass path blocks are also not rendered.

NoxNebula:
Tested the current master, rendering works with 1.9 textures, but pathblocks are still missing :(

CounterPillow:
Path blocks are implemented in #1263 but somebody needs to review and merge that.

eminence:
thanks for digging up that PR, i'll take a look tonight

eminence:
Path blocks are now added in 5f9276a24a8d64efe4b91cf29c565b4541a0cdf0

Still todo is check on command blocks

jason-green-io:
Chorus plants too :)

eminence:
those are weiirrddd.  but i think we can do them

NoxNebula:
The command blocks are not rendered correctly (see the arrows) and the cained and the repeating are not rendered at all: http://imgur.com/a/jDs64

eminence:
looks like there are two new variants of command blocks that need to be added

tazorax:
Beetroot crops added on PR #1284

[...]

PR #1283, opened 2016-03-05
Fix and add some  block for Minecraft 1.9

Lother:
first time to create a pull request :)

- **end gateway** look like a block of end portal
- New crops **beetroot**
- Fix block orientation of **purpur pilar**
- Add **frosted ice**, **structure block**
- Add **repeating command block** and **chain command block** ,
Command block need orientation data, i dont know how to got it. :P

Not fix. I don't know how to draw it  :P
ID	Name
198	end_rod
199	chorus_plant
200	chorus_flower

### DIFF
--- a/overviewer_core/textures.py
+++ b/overviewer_core/textures.py
@@ -4017,11 +4017,20 @@ def cocoa_plant(self, blockid, data):
     return img
 
 # command block
-@material(blockid=137, solid=True, nodata=True)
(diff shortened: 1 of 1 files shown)


eminence:
CC #1280

tazorax:
Beetroot crops have only data between 0 to 3, not the same as carrots and potatoes (0-7).
See PR #1284

Issue #992, opened 2013-09-22
Gap between stairs


alexschrod:
Unlike most other blocks, it seems that Overviewer is having trouble stitching together stairs, as you can see in this picture:
![stair-gap](https://f.cloud.github.com/assets/767490/1188362/30190f52-23ca-11e3-83c0-19c1e01b7e88.png)

I'm using Overviewer 0.10.92 for this render.

eminence:
Closed by #1285 !

NoxNebula:
It's not fixed, slabs still have this issue, and they look weird:
![Slabs](http://i.imgur.com/lYOFmgJ.png)
These are all sandstone slabs, but double slabs (full blocks) are lower

Socolin:
For slab see #1286

NoxNebula:
thanks

PR #1285, opened 2016-03-05
Add some missing pixel between blocks (Issue #992)

Socolin:
Just copy/paste this piece of code from `build_block` to `build_full_block`
It fix issue #992


eminence:
nice work.  thanks!

PR #1286, opened 2016-03-06
Add missing pixels between slabs

Socolin:
Like stairs in issue #992 slabs have same problem


agrif:
Thanks for doing this pixel-fiddling for us, I know `textures.py` is a nightmare. Merged!

PR #1288, opened 2016-03-06
Chorus plants and chorus flowers support

tazorax:
Add chorus plants and chorus flowers for issue #1280


agrif:
A bit late, but thank you very much for the addition!

PR #1307, opened 2016-06-10
add new 1.10 blocks

dbergl:
Add Magma block, Nether wart block, Red nether brick, and Bone block


agrif:
Awesome, merged, thank you for your help!

PR #1311, opened 2016-06-28
Leaflet

eminence:
Do not merge yet.  I opened this PR just so people can track status

Things to do:
- [x] Add a world switcher control
- [x] re-add support for overlays
- [x] re-add support for updating the url hash (so you can copy/paste a URL to link to a specific location)
- [x] I _think_ we still use a few small assets from gmaps?  Check if we do and if so, remove/replace them
- [x] re-add support for showing the chunk/region coords
- [ ] ... what else?

Discussion points:
- We are _totally_ dropping support for the google maps API.  Some of our users might have added some customized stuff to their maps, which will require updating.   In theory, we could support both google maps and leaflet, but I'm not sure I'm in favor of this, just because of the extra work involved.  If anyone has opinions on this, please leave a comment

PR #1315, opened 2016-07-21
Add support for Google Maps API key

mikewebkist:
The Google Maps JS API now requires an API key. This adds support for a "google_api_key" variable in the config file, which is then properly added to the index.html when it is generated.


CounterPillow:
We're currently working towards porting everything to LeafletJS in #1311, so this gets a NACK for me since I hope we can definitely and finally drop Google Maps API very soon.

KrzysztofJaneczko:
Could you please merge these changes? I'm surprised you rejected them in the first place.

CounterPillow:
@KrzysztofJaneczko because I'm not sure how this will interact with local map viewing or domains that have been grandfathered in, and I'd rather have a breaking change be something that gets rid of Google Maps.

rlightner:
Would love to see this merged in since work on Leaflet seems to be dead.

monokrome:
This seems pretty cool. Any chance it'll ever be merged? or is there possibly a new solution to this problem since it has been quite a while?

CounterPillow:
Google Maps API will be dead this year due to a requirement for a credit card, so we're switching to leaflet, which means this isn't needed.

monokrome:
@CounterPillow How does a credit card requirement make it dead? I would be fine w/ that.

CounterPillow:
@monokrome most people are not happy providing their credit card details just for loading one JS file that we are not allowed to distribute ourselves. We used a very small part of the Google Maps API, essentially just the JS components, and yet our users would've had to sign up for services as if they were going to run a business drawing things on top of Google's tiles.

Leaflet is a better solution. It doesn't require an account with a surveillance megacorporation to use it. It doesn't require our users to give said corporation their credit card details for a free service. We can distribute it ourselves. We can modify it to our liking. We can pin it to a specific version. We can use it without an internet connection.

monokrome:
To be clear, I'm not trying to be a proponent of Google Maps here. I'm just trying to get this thing to work without needing a secondary script injecting API keys whenever files change. If the change to leaflet has made that possible, then I'm excited 😺

PR #1319, opened 2016-08-09
Update text for 1.10

manuelgu:
Updated the text for getting the latest minecraft .jar to be for 1.10, used to be 1.8.


manuelgu:
Thinking about it, it could actually be updated to use `1.10.2` in general.

Issue #1271, opened 2016-01-18
Error running 0.12.87 with --genpoi


Bobcat00:
Windows 8.1 64-bit overviewer-0.12.87.zip gets the following error with --genpoi:

```
C:\Users\Bob\Documents\Overviewer>overviewer.exe --config=worlds/config.py --genpoi --skip-scan
2016-01-17 21:29:25  Loaded UUID cache from 'C:\\users\\bob\\documents\\overviewer\\maps\\uuidcache.dat' with 0 entries
2016-01-17 21:29:26  Done handling POIs
2016-01-17 21:29:26  Writing out javascript files
2016-01-17 21:29:26  Wrote UUID cache with 0 entries
2016-01-17 21:29:26 E An error has occurred. This may be a bug. Please let us know!
See http://docs.overviewer.org/en/latest/index.html#help

This is the error that occurred:
Traceback (most recent call last):
  File "overviewer.py", line 605, in <module>
  File "overviewer.py", line 130, in main
  File "overviewer_core\aux_files\genPOI.pyc", line 533, in main
  File "overviewer_core\aux_files\genPOI.pyc", line 241, in save_cache
  File "overviewer_core\files.pyc", line 164, in __exit__
WindowsError: [Error 32] The process cannot access the file because it is being used by another process
```

The following is the output with version 0.12.82:

```
C:\Users\Bob\Documents\Overviewer>overviewer.exe --config=worlds/config.py --genpoi --skip-scan
2016-01-17 21:32:15  Loaded UUID cache from 'C:\\users\\bob\\documents\\overview
er\\maps\\uuidcache.dat' with 0 entries
2016-01-17 21:32:16  Done handling POIs
2016-01-17 21:32:16  Writing out javascript files
2016-01-17 21:32:16  Wrote UUID cache with 0 entries
2016-01-17 21:32:16  Done
```

eminence:
Perhaps another instance of overviewer is running?  Something is holding open the yuid cache file

Bobcat00:
Definitely only one instance running. And I immediately reverted to 0.12.82 and re-ran Overviewer in the same command window with no problems.  .82 to .87 involved changing the cache processing, so maybe that's where the problem lies.

CounterPillow:
I just pushed a change which may fix this, let us know if the problem still persists.

Thanks for your bug report.

Bobcat00:
Nope, running Win 64-bit 12.90 gets the following error:

```
[...]

PR #1321, opened 2016-08-17
Fix for #1271: Delete rename destination so Windows doesn't throw an Error

calebsander:
I have tested this and it fixes the issue for me; I no longer have to delete `uuidcache.dat` before running with `--genpoi`.


CounterPillow:
I'd rather have this be specifically only doing it on Windows, since otherwise a lot of unnecessary file system operations will be done on Linux.

More importantly, I don't get why we use FileReplacer in genpoi in the first place. Can't it just always overwrite the cache file?

calebsander:
Okay, fixed it so it only does the remove on Windows. I don't know why `FileReplacer` is used; I was trying to make the program run, not redesign it.

CounterPillow:
The problem was that capabilities weren't queried properly, the FileReplacer thing works properly when fs caps are known.

Issue #1332, opened 2016-10-05
2D Overlays on a 3D World


Xeyler:
Hey, I'm working on a project called BitQuest. You can find it's page [here](https://github.com/bitquest/bitquest). We have player land claims in the size of chunks(16 blocks * 16 blocks), and we have recently been discussing adding overlays to our map that correspond to each land claim(you can see the discussion [here](https://github.com/bitquest/bitquest/issues/97)).

I am worried about how the overlays will line up with the chunks, though. If each chunk can have blocks with varying Y coordinates, the isometric overlay will not be able to accurately represent where each chunk is. I suspect overlays are flat surfaces displayed at a height of 64 blocks, but if there's a chunk with blocks at the height 256, the overlay won't accurately line up with it.

In this [example](https://cloud.githubusercontent.com/assets/11788703/19130159/54da4926-8b07-11e6-82c3-591fae96af43.png), you can see how it might be difficult to know the exact bounds of the territories because the ground and the overlays don't match up exactly thanks to varying block heights, but with chunk-sized overlays, I'm afraid it could be a lot more obvious and problematic.

Is there any way we can render the map with a top view instead of an isometric one, or any other workaround for representing 2D overlays in a 3D world?

CounterPillow:
Our overlays already draw on top of the highest block. See examples such as the slime overlay.

![](https://pbs.twimg.com/media/CmLbOjAXIAAH0p5.jpg:large)

>Is there any way we can render the map with a top view instead of an isometric one

no

PR #1336, opened 2016-10-23
Added BoundsOverlay Rendermode Primitive

jpchip:
As a response to #1332 and designed to [Display land ownership on map.bitquest.co](https://github.com/bitquest/bitquest/issues/97), I created a new RenderMode Primitive called `BoundsOverlay`. This rendermode overlays colors on the map within sets of bounding boxes, specified as multiple tuples of the form `(minx, minz, maxx, maxz)`. So by specifying `(0, 0, 16, 16)` all the blocks (inclusive) within the boundary created by these coordinates will be colored. The color is specified as `(r, g, b, a)`. 

An example usage in the config.py would be:

```
renders["boundsoverlay"] = {
        'world': 'woot',
        'title': 'Land Ownership',
        'rendermode': [ClearBase(), BoundsOverlay(bounds=[
                        (((0, 0, 16, 16), (-70, -30, -54, -14)), (255, 160, 122, 127)),
                        (((100, 100, 116, 116),(17, 0, 33, 16)), (75, 0, 130, 127)),
                       ])],
        'overlay': ['day']
}
```

with a result like:

screenshot 2016-10-20 21 26 19

I also updated the docs with example usage.

### DIFF
--- a/overviewer_core/src/primitives/overlay-bounds.c
+++ b/overviewer_core/src/primitives/overlay-bounds.c
@@ -1,0 +1,249 @@
+/*
+ * This file is part of the Minecraft Overviewer.
+ *
+ * Minecraft Overviewer is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Minecraft Overviewer is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with the Overviewer.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "overlay.h"
+
+typedef enum { false, true } bool;
+
+typedef struct {
+    /* inherits from overlay */
+    RenderPrimitiveOverlay parent;
+    void *bounds;
+    int numcolors;
+} RenderPrimitiveBounds;
+
+struct Condition{
+    int minx, minz, maxx, maxz;
+};
+
+
+struct Color {
+    int numconds;
+    struct Condition *conditions;
+    unsigned char r, g, b, a;
+};
+
+/*
+* Determines if point x,z is inside bounds minx, minz, maxx, maxz
+*/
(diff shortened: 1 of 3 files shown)


overviewer_core/src/primitives/overlay-bounds.c
```diff
@@ -0,0 +1,249 @@
...
+    struct Condition * c = NULL;
+    bool any = true;
+	
+    /**
+     * Check for every color in the current point is in the given bounds,
+	 * and color appropriately
+     **/
+    // iterate over all the colors
+    for ( col = 0; col < self->numcolors; col++) {
+		any = false;
```
gluth: Code format need more care here and over the whole PR.

overviewer_core/src/primitives/overlay-bounds.c
```diff
@@ -0,0 +1,249 @@
...
+    // opt is a borrowed reference.  do not deref
+    // store the bounds python object into opt.
+    if (!render_mode_parse_option(support, "bounds", "O", &(opt))) {
+        return 1;	
+	}
+
+    /**
+     * Check if a sane option was passed.
+     **/
+    if (opt && opt != Py_None) {
```
gluth: This `if` is taking whole function with exception for one command at the end. I would invert it and close function fast. By example:

```
if (!opt || opt == Py_None) {
  self->parent.get_color = get_color;
  return 0;
}
```

There is one line which will be repeated, but overall this will increase code readability with limited number of levels.

overviewer_core/src/primitives/overlay-bounds.c
```diff
@@ -0,0 +1,249 @@
...
+        self->numcolors = bounds_size;
+        if (bounds == NULL) {
+            PyErr_SetString(PyExc_MemoryError, "failed to allocate memory");
+            return 1;
+        }
+
+        /**
+         * Try to parse the definitions of conditions and colors.
+         **/
+        if (cont) {
```
gluth: Similar situation here, if not cond set parent color and return 0. There is no need to wait with this till the function end.

overviewer_core/src/primitives/overlay-bounds.c
```diff
@@ -0,0 +1,249 @@
...
+        // freeing the nested bounds
+        struct Color * m = self->bounds;
+        for(i = 0; i < self->numcolors; i++){
+            if(m[i].conditions) {
+				free(m[i].conditions);
+			}
+        }
+    }
+
+    if (self->bounds) {
```
gluth: The same condition in two ifs side by side, join them into one if or fail fast if there is no `self->bounds`.

PR #1343, opened 2016-11-22
Updating linux build docs

eminence:
Include some small text that clarifies how to run overviewer after a `build` step

CC @agrif @CounterPillow @rrimc69

Does this help reduce some of the confusion?

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -133,6 +133,8 @@ You will need the following packages (at least):
 Then to build::
 
     python setup.py build
+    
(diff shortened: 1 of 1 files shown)


CounterPillow:
`setup.py install` doesn't just install "things" into `/usr/local/bin/`, it also throws the python modules into `/usr/local/share/` IIRC. As it stands, people might think that by removing overviewer.py from `/usr/local/bin/` they've gotten rid of the local installation, when the modules in the local python path will still cause issues.

eminence:
What about:

> Note that there is a `python setup.py install` step that you can run which will install things into ``/usr/local/bin`` and other system-wide locations, but this is not recommended as it might conflict with other installs of Overviewer

CounterPillow:
Change "might" to "will", and put emphasis on "not recommended".

agrif:
maybe: **strongly not recommended**. The issues this caused when we last had it in our documentation were subtle and common.

CounterPillow:
What if we just leave away `setup.py install` altogether and only tell people to add it to their `$PATH` instead?

Also, change "python" to "python2" if you leave it in.

PR #1346, opened 2016-11-29
1.11 blocks

Rycieos:
Add new blocks in 1.11, the observer and shulker box. Both have been tested and compared to in game. Test render can be seen [here on my dev server](http://devmaps.rycieos.com/#/779/64/395/max/0/0).

The observer was straightforward like a dispenser, but the shulker box was much more complex. The shulker box does not have any block textures, instead it is rendered from the shulker mob texture. This meant I had to hack the parts from the mob texture to get it to render correctly.


CounterPillow:
Seems fine, merged. Thanks!

PR #1359, opened 2017-01-11
Removing deprecated option in index.html : ?sensor=false

mioux:


agrif:
I've merged this, sorry for the wait!

PR #1400, opened 2017-07-16
Fix slime chunk algorithm

lyknode:
Hi,

I discovered that the SlimeOverlay gives slightly different results that the Slime Finder app. After comparing the algorithm used and doing an in game test, it appears that the SlimeOverlay is wrong on some random chunk.

This is due to the original in game implementation that use an int as chunk coordinates. As a result, sometime an int overflow occurs and is taken into account into the final result, which does not happen in the SlimeOverlay.

I made a correction and tested it. It now displays slime chunk correctly.


agrif:
Merged, thank you! Sorry for leaving this hanging so long.

Issue #1406, opened 2017-08-22
Glazed Terracotta appears to need orientation handling


bobfrankly:
**The Issue:**
Currently Overviewer does not appear to have orientation handling for Glazed Terracotta, as all GT blocks currently display the exact same texture regardless of orientation. Proper orientation rendering is how all those neat patterns occur, so here's the issue report. :smile: 

**The Explanation:**
_Full Disclaimer: First time trying to read python. Please advise if I'm off my rocker._
Glancing through the textures.py, it looks as I thought it might. Block orientation isn't even checked for unless specified, It looks like logs [have an orientation handling](https://github.com/overviewer/Minecraft-Overviewer/blob/39d588ae45d5e66f92e2c7b8923e5269e601f3b5/overviewer_core/textures.py#L1122), however glazed terracotta [does not](https://github.com/overviewer/Minecraft-Overviewer/blob/39d588ae45d5e66f92e2c7b8923e5269e601f3b5/overviewer_core/textures.py#L4680).

Glazed Terracotta displays it's textures differently, depending on the direction the block is facing, similarly to logs, but potentially with more detail (_north and south facing logs likely look largely the same, whereas north and south facing glazed terracotta do not_), so I suspect new code will need to be authored for G.T. blocks, or I would have dragged and dropped from logs or something similar.

bobfrankly:
Adding some pics that should help make the issue more clear:
**In game:**
![2017-08-22_13 58 50](https://user-images.githubusercontent.com/2328006/29587463-f2373da4-8742-11e7-8f1d-efa969fe602d.png)
**Vs Overviewer Render:**
![capture](https://user-images.githubusercontent.com/2328006/29587881-37a7184a-8744-11e7-9b7b-726ec242599f.PNG)


Also, a reference shot that may be of assistance. Blocks are arranged in a compass formation, diamond indicating the north. Directions in this reference are according to the direction the PLAYER was facing, so they may be inverted according to [this](https://minecraft.gamepedia.com/Glazed_Terracotta#Block_data).
![2017-08-22_14 12 59](https://user-images.githubusercontent.com/2328006/29587825-06133f7a-8744-11e7-9e1a-b04c7e4f6108.png)

bobfrankly:
PR #1408 Should resolve this once approved.

bobfrankly:
PR #1408 resolved issue.

PR #1408, opened 2017-08-24
Terracotta texture rotation fix

bobfrankly:
Terracotta blocks have differing texture rotations on each side, necessitating different handling for each directional map render. They also have different rotations for each of their 4 possible orientations.

This is a fix for  #1406. It was aligned against "Light Grey Glazed Terracotta", I'm presuming the rest of the colors follow.


bobfrankly:
Overviewer results after applying this pull request:
ov-gtfixed

Ingame comparison - matching.

PR #1456, opened 2018-05-20
[WEB] JS Fixes

DevScyu:
Added support for Polygons, Polylines and Fixed a bug where you could scroll to a minimum zoom level of 0, even if in the config it was set at a higher number.

Fixed the code for 
```js
overviewer.util.ready(function(){
    // do stuff
});
```

See the changes in action: http://wynndata.tk/map/

### DIFF
--- a/overviewer_core/aux_files/genPOI.py
+++ b/overviewer_core/aux_files/genPOI.py
@@ -385,8 +385,6 @@ def create_marker_from_filter_result(poi, result):
     # read some Defaults from POI
     if "icon" in poi:
         d["icon"] = poi['icon']
-    if "createInfoWindow" in poi:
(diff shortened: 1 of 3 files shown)


r15ch13:
👍 for the  `overviewer.util.ready();` fix

CounterPillow:
Sorry that I haven't gotten around to reviewing this yet, I'll try to get it done within the next 7 days.

DevScyu:
I have been working on the Leaflet map for a while so any questions or changes feel free to ask.

SkylixMC:
In reference to the bug where you could zoom past the number set in the config, though now it is fixed for the most part, still if you switch to another world, the world loads so it IS zoomed past the min zoom. You have to zoom back in to fix it.

PR #1460, opened 2018-06-13
Remove jQuery and underscore dependencies

r15ch13:
- replace `$.each` with [`.forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)
- replace `_.each` with `.forEach`
- replace `_.isFunction` with a `typeof` comparison


CounterPillow:
Thanks!

PR #1468, opened 2018-07-28
Missing underscore.js/backbone.js in index.html

cvsync:
Overviewer depends on Backbone framework, but missing to load underscore.js/backbone.js in index.html.


CounterPillow:
We no longer depend on these, I think

cvsync:
Hmmm. By my browser devtools (chrome/firefox) warns that, and overviewer doesn't work. Anyway, I'll retry to reinstall overviewer from scratch. Thanks.

Issue #1403, opened 2017-08-02
Running Overviewer on Win10 and MC vanilla 1.12 and get an error


drysz:
I had run overviewer on this Win10 machine back in February but I upgraded to MC 1.12. It didn't work so I went and downloaded overviewer-win64-0.12.196.zip. I deleted my c:\mcmap contents and tried running it again with the same error. any ideas on how to fix this? Thanks! This is Minecraft server running on Win10.

~Dave / H0logram

C:\Minecraft\OverViewer>c:\minecraft\overviewer\overviewer.exe --verbose C:\Minecraft\RyszCraft2\RyszCraft c:\mcmap
overviewer.py:275  5264 2017-08-02 18:52:59 DEBUG    Using 'C:\\Minecraft\\RyszCraft2\\RyszCraft' as the world directory
overviewer.py:276  5264 2017-08-02 18:52:59 DEBUG    Using 'c:\\mcmap' as the output directory
overviewer.py:342  5264 2017-08-02 18:52:59 INFO     Welcome to Minecraft Overviewer!
overviewer.py:343  5264 2017-08-02 18:52:59 DEBUG    Current log level: 10
files.py:43        5264 2017-08-02 18:52:59 DEBUG    Detected that chmods work in 'c:\\mcmap'
files.py:56        5264 2017-08-02 18:52:59 DEBUG    Detected that overwriting renames do NOT work in 'c:\\mcmap'
assetmanager.py:57 5264 2017-08-02 18:52:59 DEBUG    Traceback (most recent call last):
  File "overviewer_core\assetmanager.py", line 51, in __init__
IOError: [Errno 2] No such file or directory: 'c:\\mcmap\\overviewerConfig.js'

assetmanager.py:68 5264 2017-08-02 18:52:59 DEBUG    Preferred enoding set to: cp1252
[...]

Issue #1312, opened 2016-06-30
Error -3 : Invalid literal/lengths set


mckerrnel:
This is being run against a 1.9.4 map.  I don't know what this error is...   I was running this fine several months ago against 1.8.9 map.  I did install the newest version of Overviewer.

2016-06-30 15:50:03  Welcome to Minecraft Overviewer!
2016-06-30 15:50:03  Forcerender mode activated. ALL tiles will be rendered
2016-06-30 15:50:03 E An error has occurred. This may be a bug. Please let us know!
See http://docs.overviewer.org/en/latest/index.html#help

This is the error that occurred:
Traceback (most recent call last):
  File "overviewer.py", line 608, in <module>
  File "overviewer.py", line 469, in main
  File "overviewer_core\world.pyc", line 97, in __init__
  File "overviewer_core\nbt.pyc", line 29, in wrapper
  File "overviewer_core\nbt.pyc", line 37, in load
  File "overviewer_core\nbt.pyc", line 183, in read_all
  File "gzip.pyc", line 256, in read
  File "gzip.pyc", line 307, in _read
error: Error -3 while decompressing: invalid literal/lengths set

D:\Overviewer>pause & exit
Press any key to continue . . .

CounterPillow:
An early guess of mine without having looked at this deeper is that your level.dat is corrupt.

CounterPillow:
Okay, yep, it's the level.dat. Line 469 in `overviewer.py` tries to load the world, which then leads to line 97 in `world.py` which tries to load the level.dat. Can you upload your level.dat to http://overviewer.org/uploader/ so we can take a look at it?

Issue #1196, opened 2015-01-11
Problem/error on first run.


jakerules:
![image](https://cloud.githubusercontent.com/assets/6377830/5696731/2e848874-99a8-11e4-97c5-3de82121f913.png)

[15:18:43 INFO]: Current time is Jan 11, 2015 3:18:43 PM
[15:18:43 INFO]: System.getProperty('os.name') == 'Windows 7'
[15:18:43 INFO]: System.getProperty('os.version') == '6.1'
[15:18:43 INFO]: System.getProperty('os.arch') == 'amd64'
[15:18:43 INFO]: System.getProperty('java.version') == '1.7.0_71'
[15:18:43 INFO]: System.getProperty('java.vendor') == 'Oracle Corporation'
[15:18:43 INFO]: System.getProperty('sun.arch.data.model') == '64'

Help.....

ion1:
Do any of the directory names in .minecraft/saves have non-ASCII characters, potentially §?

CounterPillow:
I took a look into this: the directory names aren't the issue, the level names are. Minecraft doesn't let you enter `§` as a character in the level name but other non-ASCII characters such as `ä` work fine.

So I had a play with this using gzip and okteta (KDE hex editor).

```
cat level.dat | gzip -d > ayylmao
```

Changed one of the bytes in the level name to `A7`, then saved it again using

```
gzip -c ayylmao > level.dat
```

However, the output I got from Overviewer is slightly different. Notably, it crashes in the NBT loader rather than the save listing function.

```
Traceback (most recent call last):
  File "overviewer.py", line 603, in <module>
    ret = main()
  File "overviewer.py", line 227, in main
    list_worlds()
  File "overviewer.py", line 572, in list_worlds
    worlds = world.get_worlds()
  File "/home/fratti/Projekte/mco_fork/Minecraft-Overviewer/overviewer_core/world.py", line 785, in get_worlds
    info = nbt.load(world_dat)[1]
  File "/home/fratti/Projekte/mco_fork/Minecraft-Overviewer/overviewer_core/nbt.py", line 29, in wrapper
    return func(fileobj, *args)
  File "/home/fratti/Projekte/mco_fork/Minecraft-Overviewer/overviewer_core/nbt.py", line 37, in load
    return NBTFileReader(fileobj).read_all()
  File "/home/fratti/Projekte/mco_fork/Minecraft-Overviewer/overviewer_core/nbt.py", line 193, in read_all
[...]

PR #1477, opened 2018-09-07
NBT: Use a replacement strategy to deal with undecodable data.

eminence:
The NBT spec requires that all strings are UTF-8 encoded.  However in
practice, non-UTF-8 data can get into a level.dat file (generally via
some weird user-supplied characters in signs.  This results in the
following error:

    CorruptNBTError: could not parse nbt: 'utf8' codec can't decode byte
    0xc0 in position 3: invalid start byte

This happens often enough that we should just ignore this error by using
the 'replace' strategy to replace the invalid data

CC @cheeseinthedark who most recently reported this error

### DIFF
--- a/overviewer_core/nbt.py
+++ b/overviewer_core/nbt.py
@@ -144,7 +144,7 @@ class NBTFileReader(object):
         # Read the string
         string = self._file.read(length)
         # decode it and return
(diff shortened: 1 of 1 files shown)


eminence:
Closes #1312 
Closes #1403
Closes #1196

PR #1478, opened 2018-09-13
Fixed behaviour of 'defaultZoom' property.

leighleighleigh:
The initial map view was not affected by the 'defaultZoom' property in the config file. 

This has been fixed, by passing the value into 'centers' collection, when they are initially created. 

Additionally, the 'defaultZoom' config property is passed to the first overviewer.map.setView() function call, to ensure correct zoom level on the initial page load.

Note that invalid values of 'defaultZoom' (such as those above/below maxZoom/minZoom) are clamped appropriately, and do not cause any errors.


SkylixMC:
OMG that you so much <3! I have been trying to find a way to fix it, but since I do not know Python, I have been unsuccessful. Thanks soo much!

agrif:
Merged this in with #1487. Sorry for the late response!

This should fix the default zoom, at least.

PR #1487, opened 2018-10-06
Use cached center for initial view. Fixes #1453.

### LINKED ISSUE #1453: showspawn False breaks overview completely
I was wondering why newer leaflet maps were completely broken and it turns out that my config uses `showspawn` to hide the spawn icon.  

Setting `showspawn` to `False` breaks leaflet, instead of hiding the spawn icon.

14mRh4X0r:


agrif:
Merged this in with #1478. Thanks!

PR #1501, opened 2018-12-09
Fixing handling of torches/redstone torches on walls

jvaskonen:
... And other stuff if I get to it.


agrif:
Thank you for your work! I've merged this into master.

PR #1502, opened 2018-12-12
Fixing top textures of non-vertically oriented quartz pillars.

jvaskonen:
Top textures didn't agree with side textures. Side textures were correct.


agrif:
Thanks!

PR #1510, opened 2019-01-23
Fix marker icon position properly. Closes #1481.

### LINKED ISSUE #1481: Manual POI Icons positioned wrongly
Icons associated with manual POIs are anchored at the specified coordinates by their top-left corner, rather than the point at the bottom (or signpost, for the sign icon). Having this anchor point makes zooming look a little odd. Instead, I suggest that the anchoring be changed to be the same as the spawn and linked location POIs, which behave as expected.

14mRh4X0r:
CSS margin is always relative to the containing element. For some
reason, using the browser inspector, this does not get propagated
properly. Adding a containing element with the size of the child element
allows us to use margin to offset the icon.


CounterPillow:
I've tested this and can confirm that it works.

PR #1526, opened 2019-03-04
Add WebP image format support

CounterPillow:
Since Firefox 65 added support for WebP, users may be interested in having maps that use WebP images. Support for this is added in this commit, along with documentation for it.

A new option, "imglossless", controls whether we write out lossless or lossy WebP images. The generic name "imglossless" as opposed to a more specific "webplossless" was chosen in case future image formats we also implement also support lossless/lossy modes in the same format (JPEG-XL? AV1 image format?).

It's an okay meme but lossy mode really falls apart on our sorts of images on the more zoomed out composite tiles, resulting in pretty blurry messes. Might be due to a PSNR bias in the encoder, which is to be expected from Google.


CounterPillow:
Had a gander at the code again after two days, and agrif looked at it too, and neither of us saw any glaring issues so I'm merging this.

PR #1529, opened 2019-03-08
overviewer: replace optparse with argparse

CounterPillow:
optparse is deprecated, and switching to argparse is mostly but not entirely trivial. Nicely telling the user about shell quoting had to be removed because I didn't see an easy way to do that with argparse.

I'm submitting this as a PR because it is a scary change and I'd like for somebody else to test this before merging it.

### DIFF
--- a/overviewer.py
+++ b/overviewer.py
@@ -129,7 +129,21 @@ def main():
     exegroup.add_argument("--skip-players", dest="skipplayers", action="store_true",
                           help="When running GenPOI, don't scan player data.")
 
-    args = parser.parse_args()
+    args, unknowns = parser.parse_known_args()
+
+    # Check for possible shell quoting issues
+    if len(unknowns) > 0:
+        possible_mistakes = []
+        for i in xrange(len(unknowns) + 1):
+            possible_mistakes.append(" ".join([args.world, args.output] + unknowns[:i]))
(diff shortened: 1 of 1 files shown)


MasterofJOKers:
Maybe you could use [`parse_known_args()`](https://docs.python.org/3/library/argparse.html?highlight=argparser#argparse.ArgumentParser.parse_known_args) instead of `parse_args()` for the shell-quoting stuff?

CounterPillow:
Good call, I'll look into that.

CounterPillow:
Done. I haven't tried this out on Windows (where this code will get triggered the most often) so I might want to do that before merging this.

overviewer.py
```diff
@@ -129,7 +129,22 @@ def main():
...
-    args = parser.parse_args()
+    args, unknowns = parser.parse_known_args()
+
+    # Check for possible shell quoting issues
+    if len(unknowns) > 0:
+        possible_mistakes = []
+        for i in xrange(len(unknowns) + 1):
+            possible_mistakes.append(" ".join([args.world, args.output] + unknowns[:i]))
+            possible_mistakes.append(" ".join([args.output] + unknowns[:i]))
+        print(possible_mistakes)
```
MasterofJOKers: That `print()` here is a leftover, isn't it?

CounterPillow: Yikes, thanks for letting me know.

PR #1533, opened 2019-03-11
Modernise png-it.py and make it less weird

CounterPillow:
Fix up code style, grammar, port to argparse, and make it work with Python 3 (!!!). plz2test kthnx


MasterofJOKers:
If this is still supposed to work on python2, wouldn't it make sense to add an `from __future__ import print_function` at the top? I mean, you're using the print-statement with braces here and I fear someone might mistake it for the print-function and use it in python2-incompatible ways.

PR #1536, opened 2019-03-13
Support for polyline and polygon

jsmienk:
- Added support for polyline POI using Leaflet;
- Added support for polygon POI using Leaflet;
- Added two example filter functions to create polylines and polygons from signs in-game.


jsmienk:
I tried. Vim not working with me. New commits emerge...

jsmienk:
New vim and git skills acquired

CounterPillow:
Ok now you've just made the git history even worse by back-merging master instead of rebasing on top of it

jsmienk:
Okay, that's not good. I am just trying to contribute. What is the best thing to do now? Copy changes into a new branch and delete this PR, creating a new one?

PR #1539, opened 2019-03-17
Convert block ids into mc_block_id enum

Wunkolo:
At the moment, the c code is filled with conditional code and case switches using _seemingly_ magic numbers that will be very hard to maintain should there be any updates to these values with a new Minecraft version.
```c
        case 53: /* oak wood stairs */
        case 67: /* cobblestone stairs */
        case 108: /* brick stairs */
        case 109: /* stone brick stairs */
        case 114: /* nether brick stairs */
        case 128: /* sandstone stairs */
        case 134: /* spruce wood stairs */
        case 135: /* birch wood stairs */
        case 136: /* jungle wood stairs */
        case 156: /* quartz stairs */
        case 163: /* acacia wood stairs */
        case 164: /* dark wood stairs */
        case 180: /* red sandstone stairs */
        case 203: /* purpur stairs */
```
```c
                        /* doors */
                        (state.block == 64) || (state.block == 193) ||
                        (state.block == 194) || (state.block == 195) ||
                        (state.block == 196) || (state.block == 197) ||
                        (state.block == 71) || /* end doors */
```
This PR would pull all these integer values into once place with the new `mc_id.h` header that enums all of Minecraft's current item and block ids and adds the `mc_block_t` and `mc_item_t` integer alias types that can be used in place of `unsigned short` code if desired.
```c
        case block_oak_stairs:
        case block_stone_stairs:
        case block_brick_stairs:
        case block_stone_brick_stairs:
        case block_nether_brick_stairs:
        case block_sandstone_stairs:
        case block_spruce_stairs:
        case block_birch_stairs:
        case block_jungle_stairs:
        case block_quartz_stairs:
        case block_acacia_stairs:
        case block_dark_oak_stairs:
        case block_red_sandstone_stairs:
        case block_purpur_stairs:
```
```c
                        /* doors */
                        (state.block == block_wooden_door  ) ||
                        (state.block == block_iron_door    ) ||
                        (state.block == block_spruce_door  ) ||
                        (state.block == block_birch_door   ) ||
                        (state.block == block_jungle_door  ) ||
                        (state.block == block_acacia_door  ) ||
                        (state.block == block_dark_oak_door) ||
                        /* end doors */
```


CounterPillow:
Confirmed to work with our Windows builders, merged.

agrif:
This change is Good and Right but FWIW, the exact values those take on for each block are basically under our control now. Minecraft itself doesn't use explicit global ids anymore. (We still want them to be the values they are now for backwards-compatibility reasons, though.)

Wunkolo:
If that's the case then there might be a case moving forward where having some bit-tricks in the enum values (ex, having all "transparent" blocks have their last bit set or something) could speed up some places.

agrif:
That's totally doable, but would require some translation code to turn old world blockids into our internal representation. It would be nice, and also a lot of work.

CounterPillow:
Ideally we get rid of block IDs altogether and just use property bitfields, because the current 1.13 support is a huge hack that doesn't scale developer-time wise, as is textures.py.

PR #1542, opened 2019-03-19
Add .clang_format

Wunkolo:
This is to automate the styling of C code to aid contributors in making pull requests that conform to the existing style and so any future C style decisions can be made to a single file and automatically applied to the whole project.
I configured the `.clang_format` to suite most of the code found through-out the already-existing C code and based on the code review of my previous PRs.
More options can be found [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)
With this any edits to the c code can be automatically formatted by calling `clang-format` or `find . -regex '.*\.\(c\|h\)' -exec clang-format -style=file -i {} \;`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,38 @@
+---
+Language: Cpp
+AlignEscapedNewlinesLeft: 'true'
+AlignTrailingComments: 'true'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: 'false'
+AllowShortLoopsOnASingleLine: 'false'
+ColumnLimit: '0'
+ContinuationIndentWidth: '4'
+IndentWidth: '4'
+SortIncludes: true
+IncludeCategories:
+  - Regex:           '<[[:alnum:].]+>'
+    Priority:        -1
+  - Regex:           '\".*\"'
+    Priority:        1
+IndentCaseLabels: false
+MaxEmptyLinesToKeep: '1'
+PointerAlignment: Left
+SpaceBeforeAssignmentOperators: 'true'
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: 'false'
+SpacesBeforeTrailingComments: '1'
+SpacesInCStyleCastParentheses: 'false'
+SpacesInContainerLiterals: 'false'
+SpacesInParentheses: 'false'
+TabWidth: '4'
+BreakBeforeTernaryOperators: true
+Cpp11BracedListStyle: true
+BreakBeforeBraces: Custom
+BraceWrapping:
+  BeforeElse: false
+  AfterEnum: false
+  AfterFunction: false
+  AfterStruct: false
+  AfterUnion: false
+...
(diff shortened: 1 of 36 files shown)


overviewer_core/src/composite.c
```diff
@@ -26,13 +26,12 @@
 
 typedef struct {
     PyObject_HEAD
-    Imaging image;
+        Imaging image;
```
CounterPillow: wat. Am I missing something here? Why is this indented two levels deep and there's nothing after `PyObject_HEAD`?

.clang-format
```diff
@@ -0,0 +1,38 @@
...
+IndentWidth: '4'
+SortIncludes: true
+IncludeCategories:
+  - Regex:           '<[[:alnum:].]+>'
+    Priority:        -1
+  - Regex:           '\".*\"'
+    Priority:        1
+IndentCaseLabels: false
+MaxEmptyLinesToKeep: '1'
+PointerAlignment: Left
```
CounterPillow: A lot of code in Overviewer appears to use right-aligned pointers. My C knowledge isn't the best, but I think

```C
int *foo, bar;
```

makes `foo` an int pointer and `bar` an int, whereas

```C
int* foo, bar;
```

makes both of them int pointers, correct?

In that case, wouldn't right-aligned pointers be the lesser trap?

Feel free to change my opinion on this, I'm mostly just noticing a lot of `foo *bar` to `foo* bar` transformations in this PR which have me a little worried over whether this was 100% intentional.

agrif: Nope, both of those declare `foo` as a pointer and `bar` as an int. Yeah, it's pretty bad.

I used to use right-pointers because morally, I think pointerness should be part of the type. But C's syntax is such that right-pointers will get you in to trouble, like the confusion above, so nowadays I try to stick to left-pointers. I think that's a good choice for overviewer.

overviewer_core/src/composite.c
Wunkolo: It's probably due to the fact that PyObject_HEAD is a preprocessor that expands to `PyObject ob_base;` according to [this](https://docs.python.org/3/c-api/structures.html).

By adding `;` after it though it seems to fix the indent since it completes the "statement" without clang format thinking you're just using a keyword it doesn't know like putting `unsigned` and `int` on different lines.
```
typedef struct {
    PyObject_HEAD; // < adding a `;` after it fixes this.
    Imaging image;
} ImagingObject;
```

.clang-format
agrif: So: I think I swapped left and right here... 😄 

I meant, the `int *foo` style is probably more appropriate. I don't know enough about `.clang-format` to know if that's what's selected here.

Wunkolo: I'm gonna be a vote for left-pointers as defining a mix of ints and int-pointers on the same line is kind of a weird basis for having it all be on the right through-out the codebase. Would rather keep all type and type modifiers next to the type, with the name on its own(`static volatile void* pointy`).

CounterPillow:
Instead of backmerging master, can you rebase on top of it?

Wunkolo:
Cleaned the commit history up so it's just one `Add .clang_format` commit, with the current `.clang_format` revision applied.

PR #1540, opened 2019-03-17
Python 3 port

CounterPillow:
Many things work, some don't. ~~Notably, genPOI doesn't work~~, and there's some signedness comparison stuff going on in the C extension.

This also completely drops support for Python 2, as maintaining a C extension for both Python 2 and 3 is a pain and not worth it for the 9 months that Python 2 is still going to be supported upstream.

The documentation needs to be adjusted as well.

All of the few tests we have pass, and rendering a map works, both with a configuration file and without. We can also use optimizeimages.

What needs work:
- [x] documentation
  * [x] Update optimizeimages instructions
  * [x] Update build instructions
    - [x] Update Windows build instructions
    - [x] Update Linux build instructions
    - [x] Update macOS build instructions
    - [x] Update FreeBSD build instructions
- [x] genPOI
  * [x] Switch from optparse to argparse
  * [x] Update syntax and semantics to Python 3
- [x] build infrastructure
  - [x] Debian/Ubuntu
  - [ ] RHEL/Fedora/CentOS
  - [x] Windows
- [x] contrib scripts
  - [x] contributors.py
  - [x] cyrillic_convert.py
  - [x] gallery.py
  - [x] playerInspect.py
  - [x] png-it.py
  - [x] regionTrimmer.py
  - [x] rerenderBlocks.py
  - [x] testRender.py
  - [x] validateRegionFile.py

I also suggest that once this work is done, we bump the version number to 0.14 or something.

PR #1545, opened 2019-03-20
Move Contrib Scripts to Python3

SteadBytes:
`contrib` scripts to be ported or removed if no longer needed:
- [x] `contributors.py` 
- [x] `cyrillic_convert.py` 
- [x] `gallery.py` 
- [x] `playerInspect.py` 
- [x] `png-it.py`  (already done by @CounterPillow in #1540 )
- [x] `regionTrimmer.py` 
- [x] `rerenderBlocks.py` ([deleted(https://github.com/overviewer/Minecraft-Overviewer/pull/1545#issuecomment-476010679)])
- [x] `testRender.py` 
- [x] `validateRegionFile.py` ([deleted](https://github.com/overviewer/Minecraft-Overviewer/pull/1545#issuecomment-476010679))

Tests will implemented where missing
- Scripts have been refactored to achieve this easier

Scripts will be moved to `argparse` for consistency with #1540


test/test_all.py
```diff
@@ -1,26 +1,31 @@
...
 # For convenience
-import sys,os,logging
+import sys
+import os
+import logging
 sys.path.insert(0, os.getcwd())
 sys.path.insert(0, os.path.join(os.getcwd(), os.pardir))
 
-# Import unit test cases or suites here
-from test_tileobj import TileTest
```
CounterPillow: These were below the `sys.path.insert` calls for a reason; as far as I remember `sys.path.insert` alters the python module path, in this case so modules in the current working directory can be imported too. Might be relevant.

test/test_all.py
```diff
@@ -1,26 +1,31 @@
...
-from test_rendertileset import RendertileSetTest
-from test_settings import SettingsTest
-from test_tileset import TilesetTest
-from test_cache import TestLRU
-
-# DISABLE THIS BLOCK TO GET LOG OUTPUT FROM TILESET FOR DEBUGGING
+Import unit test cases or suites here
+DISABLE THIS BLOCK TO GET LOG OUTPUT FROM TILESET FOR DEBUGGING
```
CounterPillow: um...

test/test_all.py
SteadBytes: Yes I think you're right! That was `autopep8`, I should have made sure that it didn't do that - I'll change it back 👍

test/test_all.py
SteadBytes: My thoughts exactly!

SteadBytes:
@CounterPillow It looks as if some of the scripts are broken due to changes to `overviewer_core` since they were written:
- `rerenderBlocks.py` attempts to import `overviewer_core.chunk` [here](https://github.com/SteadBytes/Minecraft-Overviewer/blob/54cf6449c4bca7715afb1f12d3d1182f25ef8de5/contrib/rerenderBlocks.py#L29) which doesn't exist.
- `validateRender` calls `overview_core.nbt.load_region()` with two arguments [here](https://github.com/SteadBytes/Minecraft-Overviewer/blob/d48d1da49329e8443b36297b42ea86cb9f1e8cfe/contrib/validateRegionFile.py#L22), however that function only accepts a single argument

Is it worth me fixing these scripts as part of porting them to Python3 or are they no longer needed?

CounterPillow:
You can throw both of them out, rerenderBlocks might be useful in theory but it'd need a total rewrite to work anyway. validateRender is pre-anvil world formats, apparently, so dead and well-buried.

SteadBytes:
Ok @CounterPillow , I'll throw them out. I just realised I made a typo so fjust to be 100% sure we're on the same page - `validateRender` was meant to be `validateRegionFile`. Still ok to throw it out?

CounterPillow:
Yes, still okay to throw it out. (I checked the URL you linked and didn't even notice the typo, heh)

SteadBytes:
Hehe, I thought that was the case :smile: I'll throw it out!

SteadBytes:
Working on `testRender` unit tests to finish :+1:

CounterPillow:
Thanks, I'll try to have this reviewed either this weekend or next week, hopefully with some proposed solutions to whatever problems I see.

SteadBytes:
Thank you! I'm currently just tring to figure out some strange issues with Python 3.5 in `regionTrimmer` [here](https://github.com/SteadBytes/Minecraft-Overviewer/blob/9121a4b080fec1e85832a1df752ab0aa0d02dd76/contrib/regionTrimmer.py#L26).

The test is failing for 3.5 but not 3.6 or 3.7, I have inspected the list of nodes created in each iteration and they are the same. I can only assume its a difference in `networkx.Graph.add_edges_from` however I'm having trouble being certain :thinking:

CounterPillow:
Depending on what the underlying implementation of networkx uses, it might be related to `dict`s being unordered in 3.4 and 3.5.

SteadBytes:
Yes I think it must be something like that, I'm going to test out using `OrderedGraph` which isn't needed in 3.6+ for the exact reason you stated ([docs](https://networkx.github.io/documentation/stable/reference/classes/ordered.html)) versions. 

It is definitely `networkx`, the following snippet shows the difference:
```python
import networkx

nodes = [(0, 0), (0, -1), (-1, 0), (-1, -1)]
all_edges = [
    [],
    [],
    [
        ((-1, -1), (0, -1)),
        ((-1, -1), (-1, 0)),
        ((-1, -1), (0, 0)),
    ],
    [((-1, 0), (-1, -1))],
    [((-1, 0), (0, 0))],
    [((0, -1), (-1, -1))],
    [((0, -1), (0, 0))],
    [
        ((0, 0), (-1, 0)),
        ((0, 0), (0, -1)),
        ((0, 0), (-1, -1)),
    ],
]

graph = networkx.Graph()
graph.add_nodes_from(nodes)

for edges in all_edges:
    graph.add_edges_from(edges)

print(list(sorted(graph.edges)))
```
3.5:
```python
 [((-1, 0), (-1, -1)), ((0, -1), (-1, -1)), ((0, -1), (0, 0)), ((0, 0), (-1, -1)), ((0, 0), (-1, 0))]
```

3.6/3.7:
```
 [((-1, -1), (-1, 0)), ((-1, -1), (0, -1)), ((-1, -1), (0, 0)), ((-1, 0), (0, 0)), ((0, -1), (0, 0))]
```

SteadBytes:
No luck with `OrderedGraph`, sorting the nodes before adding edges, sorting the edges before adding them or any combination therein :sweat_smile: I have opened an [issue](https://github.com/networkx/networkx/issues/3372) on the networkx repo as I could not find any other mention of this behaviour :thinking:

test/test_regionTrimmer.py
```diff
@@ -0,0 +1,58 @@
...
+        graph.add_nodes_from([(0, 0), (0, -1), (-1, 0), (-1, -1)])
+
+        graph = region_trimmer.generate_edges(graph)
+        expected = [((-1, 0), (-1, -1)),
+                    ((0, -1), (-1, -1)),
+                    ((0, 0), (-1, -1)),
+                    ((0, 0), (-1, 0)),
+                    ((0, 0), (0, -1))]
+
+        self.assertListEqual(sorted(list(graph.edges)), expected)
```
MridulS: Hi @SteadBytes you can check for the adjacency dictionary instead, which will be less complicated than worrying about order :) 
Have a look at https://github.com/networkx/networkx/blob/master/networkx/classes/tests/test_graph.py#L591

SteadBytes: You're a :star: @MridulS that fixed it - thank you so much!

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "3.4"
```
CounterPillow: What stops 3.4 from working? I'm not too concerned about dropping support for it in the contrib scripts since it's EOL and distros still shipping it have a clear upgrade path to newer distros, but if it's something minor we might as well fix it.

contrib/regionTrimmer.py
```diff
@@ -45,114 +56,130 @@ def get_graph_bounds(graph):
...
+    return node[0] >= trim_center[0] + trim_bounds[0] or \
+        node[0] <= trim_center[0] - trim_bounds[0] or \
+        node[1] >= trim_center[1] + trim_bounds[1] or \
+        node[1] <= trim_center[1] - trim_bounds[1]
+
+def unlink_file(path):
+    try:
+        path.unlink()
+    except OSError as err:
+        logger.warn("Unable to delete file: %s", path)
```
CounterPillow: I think these here should be `logger.warning`,

```regionTrimmer.py:111: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead```

**Comment by CounterPillow** [comment]
Also, here's something to fix `contribManager.py` since it still looks for the now removed scripts:

```diff
diff --git a/contribManager.py b/contribManager.py
index 18e87a0..d3e29e1 100755
--- a/contribManager.py
+++ b/contribManager.py
@@ -10,9 +10,7 @@ import sys
 scripts = {     # keys are names, values are scripts
     "convertCyrillic":  "cyrillic_convert.py",
     "playerInspect":    "playerInspect.py",
-    "rerenderBlocks":   "rerenderBlocks.py",
     "testRender":       "testRender.py",
-    "validate":         "validateRegionFile.py",
     "pngit":            "png-it.py",
     "gallery":          "gallery.py",
     "regionTrimmer":    "regionTrimmer.py",
@@ -21,9 +19,9 @@ scripts = {     # keys are names, values are scripts
 
 # you can symlink or hardlink contribManager.py to another name to have it
 # automatically find the right script to run.  For example:
-# > ln -s contribManager.py validate.exe
-# > chmod +x validate.exe
-# > ./validate.exe -h
+# > ln -s contribManager.py pngit.exe
+# > chmod +x pngit.exe
+# > ./pngit.exe -h
 
 
 # figure out what script to execute
```

test/test_contributors.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @patch('subprocess.run')
+    def test_get_contributors(self, mock_run):
+        mock_run.return_value.stdout = dedent("""\
+            1	3decibels <3db@3decibels.net>
+            585	Aaron Griffith <aargri@gmail.com>
+            1	Aaron1011 <aa1ronham@gmail.com>
+            """).encode()
+        expected = [{"count": 1, "name": "3decibels", "email": "<3db@3decibels.net>"},
+                    {"count": 585, "name": "Aaron Griffith", "email": "<aargri@gmail.com>"},
```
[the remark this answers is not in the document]
SteadBytes: No, I mock the return value from the `subprocess.run` call here https://github.com/overviewer/Minecraft-Overviewer/pull/1545/files/b7e9e868cef080a7a7f74681c346fba236965424#diff-bd47c8cc9b35980b7699a6d6cd2c3b43R66

.travis.yml
SteadBytes: `subprocess.run` is [3.5+](https://docs.python.org/3/library/subprocess.html#subprocess.run]). I've used it to manage making external shell commands (`git` etc). You mentioned in #1540 that 3.5 was ok as a minimum - is this still the case? :smile:

contrib/regionTrimmer.py
SteadBytes: Good point - fixed it now :+1:

SteadBytes:
Thank you for the review @CounterPillow ! I have fixed the use of deprecated `logger.warn` and made the changes to `contribManager.py` (thanks for that, I totally overlooked that file :sweat: ) . I have also replied to your other comments :smile:

.travis.yml
CounterPillow: I think for contrib scripts, 3.5 is ok as a minimum. CentOS 6 still has 3.4 as the newest it can get, but I'm inclined to not really care about CentOS 6 users.

SteadBytes: Ah I see, I guess that depends how many users it would affect 🤔 Do you know how many users are on CentOS 6?

CounterPillow: I count about 70 unique IPs accessing the CentOS 6 repos in about 2 days worth of logs, but since this only affects contrib scripts which are rarely used it's practically negligible.

Admittedly, there's more CentOS 6 users than CentOS 7 users, apparently. But if they want to use the contrib scripts, they should upgrade.

SteadBytes: How interesting! CentOS 6 is eol in [November 2020](https://wiki.centos.org/About/Product) which coincides reasonably well with Python 2 eol anyway so there's some incentive to upgrade there anyway I guess. As long as the 3.5+ requirement is made clear in the documentation/help of the scripts then it should be fine? Is it worth me adding back the 3.4 environment in travis but excluding the contrib tests from it? If overviewer itself still supports 3.4 it's probably good to have the tests running for it.

CounterPillow: Disabling certain tests for certain Python versions introduces more complexity than I'd like, I think it's fine if our position on this is just "Overviewer may work with it, but if it doesn't then you should upgrade."

There really isn't any reason for them not to upgrade.

SteadBytes: That's a good point, fragmenting tests across versions is not a great idea now I think about it 😅 Sounds good, hopefully it'll help give people a nudge towards upgrading 🤞

Issue #1560, opened 2019-04-27
Missing 1.14 slabs 


IncredibleHolg:
The following slabs have to be added.

Block names and suggested block IDs:
```bash 
andesite_slab 11343
diorite_slab 11344
granite_slab 11345
polished_andesite_slab  11346
polished_diorite_slab 11347
polished_granite_slab 11348
red_nether_brick_slab 11349
smooth_sandstone_slab 11350
cut_sandstone_slab 11351
smooth_red_sandstone_slab 11352
cut_red_sandstone_slab 11353
end_stone_brick_slab 11354
mossy_cobblestone_slab 11355
mossy_stone_brick_slab 11356
smooth_quartz_slab 11357
smooth_stone_slab 11358
```

IncredibleHolg:
I have been started to implement the slabs in my fork.

IncredibleHolg:
Searched for the texture files in ```assets/minecraft/models/block```  e.g. ```smooth_red_sandstone_slab.json```states to use ```red_sandstone_top.png```.

```json
{
    "parent": "block/slab",
    "textures": {
        "bottom": "block/red_sandstone_top",
        "top": "block/red_sandstone_top",
        "side": "block/red_sandstone_top"
    }
}
```

IncredibleHolg:
![newslabs_002](https://user-images.githubusercontent.com/41533138/56861775-9b25a780-69a4-11e9-97a7-fb551dec425e.png)
![newslabs_001](https://user-images.githubusercontent.com/41533138/56861780-a8db2d00-69a4-11e9-8d5b-f7f4d0fa98a5.png)

IncredibleHolg:
Created pull request #1561

PR #1561, opened 2019-04-28
added new 1.14 slabs, referencing issue #1560

IncredibleHolg:
The new slabs are now rendring as expected.


CounterPillow:
Thanks!

Issue #1473, opened 2018-08-22
Changing Viewing Angles Messing Up Viewing Location


SkylixMC:
I rendered the same map in all 4 viewing angles (NW,NE,SW,SE).

But the problem is, when you switch your viewing angle, the focused coordinates reset themselves.

For example, if I am viewing the coordinate, x-0, y-64, z-0 from the direction North West, switching my viewing perspective to North East changing this location to x-0, y-64, z-15, and switching to South East to  x-15,y-64,z-15, and to South West to x-15, y-64, z-0.

Although I do not know how to code, I believe this issue has something to do with this line of code:
--------------------------
        // the width and height of all the highest-zoom tiles combined,
        // inverted
        var perPixel = 1.0 / (overviewerConfig.CONST.tileSize *
                Math.pow(2, zoomLevels));

        if (north_direction == overviewerConfig.CONST.UPPERRIGHT){
            temp = x;
            x = -z+15;
            z = temp;
        } else if(north_direction == overviewerConfig.CONST.LOWERRIGHT){
            x = -x+15;
            z = -z+15;
        } else if(north_direction == overviewerConfig.CONST.LOWERLEFT){
            temp = x;
            x = z;
            z = -temp+15;
        }
-----------------------------------

In addition, in the OverviewerConfig file, changing the
"UPPERLEFT": 0,
"UPPERRIGHT": 1, 
 "LOWERRIGHT": 2, 
"LOWERLEFT": 3

to

"UPPERLEFT": 0,
"UPPERRIGHT": 15, 
 "LOWERRIGHT": 15, 
"LOWERLEFT": 15

Does seem to fix this issue, and the map does stay focused to one position when rotating, BUT it messes up the coordinates so they are inaccurate (by 15 blocks).

SkylixMC:
I am checking back every few days. Will someone respond please?

SkylixMC:
bump?

PR #1571, opened 2019-05-04
Ensure coordinates are properly converted when changing views

### LINKED ISSUE #1450: Change direction/heading issue
Hi,

After changing the map heading direction, the map points to a different location (with the google maps API there was no such issue).

Example:
I generated a map heaving all directions included, but when I change map heading ("northdirection" : "upper-left", and "northdirection" : "upper-right", is also generated), the map is not having the right coordinates.
More example:
Link for the same location but changing the view from "northdirection" : "upper-left", to "northdirection" : "upper-right",:
http://link/#/619/64/956/max/Bros/Day%20-%20N
after rotate:
http://link/#/956/64/-604/max/Bros/Day%20-%20E

In practise, previously with maps API I could see the same location from different perspective.
Since Leaflet the different perspectives are pointing to different parts of the map.

Hope, it is understandable.

Minecraft Overviewer 0.12.259 (a129274)
built on Wed May  9 14:23:42 2018

Kind regards,
Dutch

mircokroon:
When changing to a layer with a different angle, the map center is not properly maintained. This PR fixes this by converting the map coordinates to world coordinates using the *current* layer's config, then converting the world coordinates to new map coordinates using the *new* layer's config. 

Closes #1450 and #1473

### DIFF
--- a/overviewer_core/data/js_src/util.js
+++ b/overviewer_core/data/js_src/util.js
@@ -236,7 +236,10 @@ overviewer.util = {
                         currentWorldCoords.z, 
                         ev.layer.tileSetConfig);
                         
-                overviewer.map.setView(newMapCoords);
+                overviewer.map.setView(
+                        newMapCoords,
+                        overviewer.map.getZoom(),
+                        { animate: false });
             }
             
             // before updating the current_layer, remove the marker control, if it exists
--- a/overviewer_core/data/js_src/util.js
+++ b/overviewer_core/data/js_src/util.js
@@ -232,8 +232,8 @@ overviewer.util = {
                     
                 const newMapCoords = overviewer.util.fromWorldToLatLng(
                         currentWorldCoords.x, 
-                        coords.y, 
-                        coords.z, 
+                        currentWorldCoords.y, 
+                        currentWorldCoords.z, 
                         ev.layer.tileSetConfig);
                         
                 overviewer.map.setView(newMapCoords);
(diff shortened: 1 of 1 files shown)


mircokroon:
I believe the pipeline failure is not related to the code I changed (?)

CounterPillow:
Yeah, no worries, that is my goof-up.

CounterPillow:
This still seems to shift the view in a strange way. To be fair though, it appears to work both ways, so that if you switch back to a non-rotated layer you do get back to the original position again, and changing to a layer with the same rotation doesn't seem to shift the view which is good.

Maybe switching to layers with rotation needs some special handling to centre onto the proper location.

mircokroon:
Hmm I wasn't 100% sure it was rotating correctly, but if I center a block at height 64 and rotate it, the same block remains centered as far as I can tell. I think that would be considered 'correct'? 

Additionally the coordinates of the map center in the URL hash remain the same, whereas they would change around previously. As these are retrieved & computed from the underlying map position I think them not being changed is a good indication of 'correct' rotation.

mircokroon:
Actually I think I may have found the issue you encountered -- when the world is near the origin (which the world I used for testing was not) then Leaflet will automatically animate changing of views, which looks bad. I've disabled that, so it should be working correctly now.

CounterPillow:
Seems to work okay enough, squashed and merged. Thanks.

PR #1588, opened 2019-06-22
[py3] Fixes for WindowsOutputStream

eminence:
I spent like 0.4 seconds on this, I'm not sure if this is the right way to do this.

cc @agrif @CounterPillow

### DIFF
--- a/overviewer_core/logger.py
+++ b/overviewer_core/logger.py
@@ -18,7 +18,7 @@ import logging
 import os
 import platform
 import sys
-from io import BytesIO
+from io import StringIO
 
 # Some cool code for colored logging:
 # For background, add 40. For foreground, add 30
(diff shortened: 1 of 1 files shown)


agrif:
This will probably fail on any string that has a unicode character in it, since you call `c.decode()` on a single byte of utf-8. Odd as it seems, you probably want to do this entirely with strings. If you don't get around to it today, maybe I will.

eminence:
This might be dumb, but what about `chr(ord(c))`

chungy:
I don't know what issue is being attempted to be resolved, but in Chocolate Doom I had to use a workaround for Unicode on Windows too: https://github.com/chocolate-doom/chocolate-doom/commit/685062a58f7be4140648dd9c7b4459e34f88cbfa

eminence:
I abandoned the encode/decode nonsense and just switched from `BytesIO` to `StringIO`

PR #1598, opened 2019-06-25
Update C codebase to standardized types

Wunkolo:
This PR is a pass across the C codebase to clean up our type usage to more concise and self-documenting [posix-standard](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdint.h.html) type aliases(such as going from `unsigned int` to `uint32_t` and `uint16_t block;` to `mc_block_t block;`).
This standardization will make further contributions to the codebase a lot easier

_(Also, in particular, being explicit with the sizes of certain types will make future data-parallelism speedups a lot easier. Like implementing a SIMD speedup when we know that a certain type will always be a `uint16_t` or `uint8_t`. Definitely been eyeing a lot of these RGB color values being split across three `uint8_t`s as function arguments)_

### DIFF
--- a/overviewer_core/src/Draw.c
+++ b/overviewer_core/src/Draw.c
@@ -71,27 +71,27 @@
 
 typedef struct {
     /* edge descriptor for polygon engine */
-    int d;
-    int x0, y0;
-    int xmin, ymin, xmax, ymax;
+    int32_t d;
+    int32_t x0, y0;
+    int32_t xmin, ymin, xmax, ymax;
     float dx;
 } Edge;
 
 static inline void
-point8(Imaging im, int x, int y, int ink) {
+point8(Imaging im, int32_t x, int32_t y, int32_t ink) {
     if (x >= 0 && x < im->xsize && y >= 0 && y < im->ysize)
         im->image8[y][x] = (UINT8)ink;
 }
 
 static inline void
-point32(Imaging im, int x, int y, int ink) {
+point32(Imaging im, int32_t x, int32_t y, int32_t ink) {
     if (x >= 0 && x < im->xsize && y >= 0 && y < im->ysize)
         im->image32[y][x] = ink;
 }
 
 static inline void
-point32rgba(Imaging im, int x, int y, int ink) {
-    unsigned int tmp1, tmp2;
+point32rgba(Imaging im, int32_t x, int32_t y, int32_t ink) {
+    uint32_t tmp1, tmp2;
 
     if (x >= 0 && x < im->xsize && y >= 0 && y < im->ysize) {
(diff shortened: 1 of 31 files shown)


Wunkolo:
Some boolean types have to stay as `int32_t` due to the usage of `render_mode_parse_option` expecting to write into an integer type(`bool`s are 1 byte in C).

Even the new `p` parsing option available since Python 3.3 still treats boolean values as 4-byte integers.
https://docs.python.org/3/c-api/arg.html

CounterPillow:
FYI I'm pretty sure we didn't use stdint.h because of fucking MSVC's lack of C99.

Wunkolo:
[Visual Studio 2013 (12.0) has `stdint.h`](https://devblogs.microsoft.com/cppblog/c99-library-support-in-visual-studio-2013/)
[Is there a MSVC version cutoff to make?](https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#History) I think it's safe to say people should at least use a version of visual studio from 6 years ago.

CounterPillow:
Does this apply to MSVC or only to MSVC++? If regular ol' C MSVC supports stdint now I'm okay with this change then.

Wunkolo:
[I'm seeing a lot of results of the introduction of `stdint.h` starting with Visual Studio 2010](https://codeyarns.com/2010/04/13/visual-studio-2010-stdint-h/) and [broader support for C99 features starting in Visual Studio 2013](https://devblogs.microsoft.com/cppblog/c99-library-support-in-visual-studio-2013/)

A little test with Visual Studio Community 2017 shows that the C compiler([Which is just `cl.exe` with the `/Tc` option. Both the C and C++ compilers are the same executable](https://docs.microsoft.com/en-us/cpp/build/reference/tc-tp-tc-tp-specify-source-file-type?view=vs-2019)) has full support for the standard integer types.

[Here's some godbolt compilation samples as well that tests some versions of the MSVC compiler for ARM, ARM64, X86, x64 across versions `19.00`(VS2015), `19.10`(VS2017),`19.14`(VS2017)](https://godbolt.org/z/WpCfFS)

agrif:
There is a very real possibility that we last visited this issue in 2010 😄 

If they've added stdint, by all means, this change is great.

CounterPillow:
I'll try to review this soon™, thanks for the work.

CounterPillow:
Tested to seemingly be working fine, merged and bumped C extension version. We really should look into those signedness compiler warnings though.

Wunkolo:
Made another PR Involving the signedness compiler warnings and others.
https://github.com/overviewer/Minecraft-Overviewer/pull/1608

PR #1608, opened 2019-07-10
Fix GCC signed-unsigned and pointer-warnings

Wunkolo:
A lot of the signed/unsigned issues are related to the fact that I converted
a lot of indexing values to use unsigned types, little did I know that a lot of
other values used when indexing actually come from the python-end of the code. Python does
not have built-in unsigned types so all integers coming from Python are signed
implicitly and a lot of things like image-size and x, y coordinates and other values coming from python come paired with conditional code to handle negative-integer cases as well.

Guess we'll just take our `int32_t i = 0; i < blah; ++i` and like it.

Also fixed some pointer-casting warnings where some functions were still returning `int32_t` where they could have just been returning the standard `bool` type

Code now compiles with no warnings or naggings of any kind on `gcc 9.1.0`.

### DIFF
--- a/overviewer_core/src/composite.c
+++ b/overviewer_core/src/composite.c
@@ -114,7 +114,7 @@ alpha_over_full(PyObject* dest, PyObject* src, PyObject* mask, float overall_alp
     int32_t sx, sy;
     /* iteration variables */
     int32_t x, y;
-	uint32_t i;
+    uint32_t i;
     /* temporary calculation variables */
     int32_t tmp1, tmp2, tmp3;
     /* integer [0, 255] version of overall_alpha */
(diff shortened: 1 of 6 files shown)


overviewer_core/src/composite.c
```diff
@@ -113,7 +113,8 @@ alpha_over_full(PyObject* dest, PyObject* src, PyObject* mask, float overall_alp
     /* source position */
     int32_t sx, sy;
     /* iteration variables */
-    uint32_t x, y, i;
+    int32_t x, y;
+	uint32_t i;
```
CounterPillow: this is a tab indent

Wunkolo: Whoops, ended up just running `clang-format` on the C code to catch this and anything like it.

PR #1619, opened 2019-08-02
 Clarified overlay transparency instructions 

Gyzie:
While configuring my own overlay I found that the docs were missing some crucial information about image transparency and dimension usage. I've updated the docs with some extra clarification about overlays.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -609,20 +609,17 @@ Rendering
 
         renders['day'] = {
             'world': 'exmaple',
-            'dimension': 'overworld',
             'rendermode': 'smooth_lighting',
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -603,17 +609,20 @@ Rendering
 
         renders['day'] = {
             'world': 'exmaple',
+            'dimension': 'overworld',
```
CounterPillow: This isn't needed, that's the default.

Gyzie: Removed in f8eba0f.

All trough I though my renders had issues when I left that out, but that's probably an issue on my part then. Might look into that again later.

PR #1628, opened 2019-08-16
Fixed sapling textures and added bamboo sapling

Gyzie:
Saplings in current master:
afbeelding

Sapling in this pull request:
afbeelding

### DIFF
--- a/overviewer_core/textures.py
+++ b/overviewer_core/textures.py
@@ -970,21 +970,23 @@ def wooden_planks(self, blockid, data):
     if data == 5: # dark oak
         return self.build_block(self.load_image_texture("assets/minecraft/textures/block/dark_oak_planks.png"),self.load_image_texture("assets/minecraft/textures/block/dark_oak_planks.png"))
 
-@material(blockid=6, data=list(range(16)), transparent=True)
+@material(blockid=6, data=list(range(19)), transparent=True)
 def saplings(self, blockid, data):
     # usual saplings
     tex = self.load_image_texture("assets/minecraft/textures/block/oak_sapling.png")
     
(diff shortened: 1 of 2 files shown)


overviewer_core/textures.py
```diff
@@ -975,16 +975,18 @@ def saplings(self, blockid, data):
     # usual saplings
     tex = self.load_image_texture("assets/minecraft/textures/block/oak_sapling.png")
     
-    if data & 0x3 == 1: # spruce sapling
+    if data == 1: # spruce sapling
```
CounterPillow: pretty sure that `& 0x3` was there because pre-1.13 Minecraft stored growth age in the block data. I propose we investigate how acacia saplings were stored in 1.12, and then fix the function for that, and then use separate IDs for the new ones and the `sprite` shorthand for them instead of using the data field and having a needlessly complicated material definition.

Gyzie: I did some quick poking about and the current master does not even work below 1.13:
afbeelding
afbeelding

All the sapling blocks pre-1.13 are just called `minecraft:sapling` with a `type` tag. And to be frank I don't see why that one oak sapling actual rendered since I can't find a reference to `minecraft:sapling`. (I guess it still works via an id in this version? Is has been too long 😅)

So I guess the best option would be create separate sapling blocks with their own ID and add an issue for sapling backwards compatibility?

CounterPillow: >So I guess the best option would be create separate sapling blocks with their own ID and add an issue for sapling backwards compatibility?

Probably, yeah.

Gyzie:
I've reworked these changes as #1634

Issue #1613, opened 2019-07-18
Missing 1.14 blocks


Clayn:
Similar to #1560 the stairs for new slabs are missing.
Also Bamboo is missing in the overviewer map.  
In this image the roof is missing the red nether brick stairs

<img width="813" alt="grafik" src="https://user-images.githubusercontent.com/8006856/61439101-3180be00-a941-11e9-9d87-95850b938135.png">

IncredibleHolg:
Stairs should be easy to fix, sonce the textures are available.

Bamboo is someting different, since the texture is generated in game from  a json model using the multipart texture file. Similar problem as #1569

IncredibleHolg:
@Clayn Can you deliver a list containing the names of the missing blocks?

Clayn:
@IncredibleHolg Sure i will during the day.

Clayn:
Missing or Wrong Blocks (in the order they appear in the test map from west to east)

1. Spruce Sign 
2. Birch Sign
3. Jungle Sign
4. Acacia Sign
5. Dark Oak Sign
6. Stone Stairs and Slabs (seems they just use the wron texture=
7. Brick Wall
10. Granite Stairs & Wall
11. Polished granite Stairs
12. Diorite Stairs & Wall
13. Polished diorite Stairs
14. Andesite Wall
15. Polished andesite Stairs
16. Sandstone Wall
17. Red sandstone Wall
18. Smooth Quarz Stairs
19. Prismarin Walls
22. Netherbrick Walls
23. Red Netherbrick Stairs & Walls
24. Endstone Brick Stairs & Wallds
26. Barrels (different directions)
27. Scaffolding
28. Bell
29. Campfire
30. Composter
31. Lantern
32. Lectern (using bottom texture of the bookshelf i guess)
34. Grindstone (rendered incorrectly)
36. Stonecutter different directions (only rendering the bottom half)
38. Bamboo
39. Cornflower
40. Wither Rose
41. Lilly of the Valley

![grafik](https://user-images.githubusercontent.com/8006856/62134809-71e92000-b2e1-11e9-8d3b-718c885028e9.png)


![2019-07-30_15 46 37](https://user-images.githubusercontent.com/8006856/62134772-672e8b00-b2e1-11e9-9378-efc353024fc6.png)
[...]

PR #1633, opened 2019-08-21
add new stone walls

IncredibleHolg:
* reordered IDs to get one numer block for the walls
* implemented new walls


IncredibleHolg:
Addresses #1613

PR #1634, opened 2019-08-21
Created seperate sapling blocks and added bamboo_sapling

Gyzie:
I've reworked the changes I made in #1628 

This PR adds bamboo again but leaves the original sapling code for backwards compatibility. This means that a bug that renders the dark_oak_sapling and acacia_sapling incorrectly is still active (I will create a bug report for this).

Render for 1.2.1:
afbeelding

1.13:
afbeelding

1.14 (current):
afbeelding


CounterPillow:
#TeamTrees

PR #1640, opened 2019-08-31
feat: added 1.14 scaffolding block

Dravere:
As the title says I've added the scaffolding block to Overviewer. Just the top texture with the side texture, transparent, non-solid.

Result:
scaffolding-overviewer

First time working on the Overviewer code. I hope I did everything right?


CounterPillow:
Thanks, I see IncredibleHolg included this in his PR, so it's merged as 935bdb847d00398de625da6131ef3006b08fa5a0

PR #1652, opened 2019-09-29
Docs: Add mkdir command to one liner because the destination forlder does not exist yet.

SkullMasher:
Add mkdir command to one liner because the destination forlder does not exist yet.

Possible alternative with curl could be `curl https://overviewer.org/textures/1.14 --create-dirs -o ~/.minecraft/versions/1.14/1.14.jar` (not tested)


SkullMasher:
Closed due to redundency with #1645

Issue #1551, opened 2019-04-15
[1.13] Nether Lava-NetherRack Pillars


durand1w:
For 1.13 on Linux, I'm using version .73 of the Overviewer.  There is a rendering issue where it is showing all the hidden lava blocks and then rendering nether rack from each one to the floor.  Example can be seen in the world map below.  This was not an issue on for 1.12.

[Nether Map](http://livemap.ambethtyr.com/#/-249/64/242/-1/world%20-%20nether/ambethtyrnether)

n4ru:
Still occurs in 1.14

supergnaw:
It's June and this is still an issue.

CounterPillow:
@supergnaw so are you gonna fix it?

supergnaw:
Where in the code is the nether render mode? If I can look at how it's detecting the ceiling maybe I can modify it to ignore a single lava block that is surrounded netherrack.

CounterPillow:
https://github.com/overviewer/Minecraft-Overviewer/blob/master/overviewer_core/src/primitives/nether.c

supergnaw:
It clearly states block_lava so now I'm not sure what could be causing this. Is it possible that block_lava isn't actually the correct block? This requires further investigation because now it's gonna bug me.

robosphinx:
The Flattening (1.13+) seems to have renamed all lava variants (flowing/stationary) to "lava"

supergnaw:
so [the namespace id is supposedly just "lava"](https://minecraft.gamepedia.com/Lava/ID) rather than block_lava, but the roof is still removed with other items in the list appended to "block_" eg block_quartz_ore or block_bedrock. where is overviewer declaring these namespaces locally? maybe that's the source of the issue.

robosphinx:
Perhaps there's a relation between [the ID assigned in blockmap in world.py](https://github.com/overviewer/Minecraft-Overviewer/blob/522ee282192276dad70b1f00de7701db8f7fd105/overviewer_core/world.py#L325) (10) and [the ID in mc_id.h](https://github.com/overviewer/Minecraft-Overviewer/blob/522ee282192276dad70b1f00de7701db8f7fd105/overviewer_core/src/mc_id.h#L19) (11)?

drnixon:
Changing from 10 to 11 in world.py and rebuilding Overviewer seems to fix the hanging lava block issue. At least for the render I just ran.

drnixon:
[...]

PR #1670, opened 2019-11-01
Change lava ID from 10 to 11 (Fixes issue #1551)

robosphinx:
Hiding the nether ceiling causes issues in 1.13+. Any block of lava in the ceiling and all netherrack directly underneath would not be hidden properly. Not tested on 1.12 or below.

### DIFF
--- a/overviewer_core/world.py
+++ b/overviewer_core/world.py
@@ -317,7 +317,7 @@ class RegionSet(object):
             'minecraft:sapling': (6, 0),
             'minecraft:bedrock': (7, 0),
             'minecraft:water': (8, 0),
-            'minecraft:lava': (10, 0),
+            'minecraft:lava': (11, 0),
             'minecraft:sand': (12, 0),
             'minecraft:red_sand': (12, 1),
             'minecraft:gravel': (13, 0),


Gyzie:
Good find, so the block ID just got translated wrongly. [This commit](https://github.com/overviewer/Minecraft-Overviewer/commit/50f12827f56e9be155962cdaa36d8c28a30ddd87#diff-550bbb85bacbfb099a331fa4a3065e85) proves that. So this should work on older versions of Minecraft.

If this isn't enough prove, and I can find the time, I might test this commit on an older world.

agrif:
ID 10 is "flowing lava" `minecraft:flowing_lava`, and ID 11 is "still lava" `minecraft:lava`. I don't remember what the difference is, but at least in that commit we only use the still version. Maybe we should check if we use flowing lava anywhere.

Gyzie:
I've tested these changes in version 1.15.1 and 1.10.2 and found no issues with changing the lava ID:

1.15.1 on current master
1_15_1-master

1.15.1 with this fix
1_15_1-fix

1.10.2 on current master
1_10_2-master

1.10.2 with this fix
1_10_2-fix

Gyzie:
@agrif I don't know if the flowing variants of lava/water were ever used in Overviewer, but currently they are not and are always rendered as a full blocks:
image

(never noticed this 😅)

Issue #1694, opened 2019-12-18
Add beehives from 1.15 to the MineralOverlay


mtoensing:
Hi! I try to find beehives in our world and would like to make them visible on a layer. So i asked myself: What is the blockid of a beehive? I look in https://github.com/overviewer/Minecraft-Overviewer/blob/master/overviewer_core/src/mc_id.h but did not find it. So how do I do it?

Gyzie:
The new 1.15 blocks have not yet been added to Overviewer and therefor don't have a block ID yet. But once they are added you will find the block ID in the file you referred to.

Gyzie:
I've created a pull request for the missing blocks #1701 once it gets merged the item id's will be:
```
block_beehive = 11501,
block_bee_nest = 11502,
block_honeycomb_block = 11503,
block_honey_block = 11504,
```

Issue #1687, opened 2019-12-11
1.15 jar not working for texture generation


ichttt:
With only the 1.15 jar present, overviewer fails with the following error:

`2019-12-11 23:16:21 E Could not find the textures while searching for 'assets/minecraft/textures/entity/chest/normal_double.png'. Try specifying the 'texturepath' option in your config file.
Set it to the path to a Minecraft Resource pack.
Alternately, install the Minecraft client (which includes textures)
Also see <http://docs.overviewer.org/en/latest/running/#installing-the-textures>
(Remember, this version of Overviewer requires a 1.14-compatible resource pack)
(Also note that I won't automatically use snapshots; you'll have to use the texturepath option to use a snapshot jar)`

If the 1.14.4 jar exists as well, the texture get generated without problems

CounterPillow:
We best start requiring a 1.15 jar once we add actual 1.15 textures I guess.

ncarroll0488:
As a really dirty trick, you can just copy in the missing 1.14 asset files into the 1.15 dir
`rsync -a --ignore-existing /path/to/1.14/ /path/to/1.15/` (trailing slash very important)
This works provided you have no new 1.15 blocks on the map. I have no idea what happens otherwise.

Gyzie:
The issue is that Mojang changed chest textures and their name (and maybe even more textures). What fix do we desire?:
- Updating the texture code so chest are only rendered correctly on 1.15 textures.
- Trying to support 1.14 and 1.15 textures by creating more complexity in the chest code.

Gyzie:
Fixed in PR #1702

PR #1701, opened 2019-12-26
Added all 1.15 bocks: bee_hive, beenest, honeycom_block and honey_bock

Gyzie:
Supports beehive/nests in all directions with or without a honey_state of 5.

image

Addresses #1694 

Mojang also decided to change the chest textures for some reason. I'll look into how we can support pre 1.15 textures and 1.15+, any suggestions? _Alright so this build breaks because of a missing 1.15 jar together with_ #1687

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,6 @@ docs/_build/
 Minecraft_Overviewer.egg-info
 terrain.png
 cachedir*
-debug_config.py
-out/
 
 #MCPatcher HD texture files
 custom_*.png
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ docs/_build/
 Minecraft_Overviewer.egg-info
 terrain.png
 cachedir*
+debug_config.py
+out/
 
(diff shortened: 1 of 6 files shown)


Gyzie:
Updated travis to Minecraft version 1.15 to resolve building issues. When #1702 is merged this build should pass.

.gitignore
```diff
@@ -6,6 +6,8 @@ docs/_build/
 Minecraft_Overviewer.egg-info
 terrain.png
 cachedir*
+debug_config.py
```
CounterPillow: please don't, you can keep those in your `.git/info/exclude` as they are specific to you

Gyzie: Reverted 👍

CounterPillow:
Thanks, I cleaned up the git history and merged it as https://github.com/overviewer/Minecraft-Overviewer/commit/885fa75f54dc5059b5085f530af259194edd62ae

(Sadly, due to the small history rewrite, github will show this as "closed" instead of "merged", but rest assured it's in with you as the author)

PR #1704, opened 2019-12-29
Fixes #1659 by testing both 8 and 9 for water blockid

### LINKED ISSUE #1659: Ocean floor covers legitimate caves when using the Cave render mode
The ocean floor blocks the view of many legitimate caves when using the Cave render mode. It would be much nicer if the ocean floor was not considered part of a cave, especially if your map is dominated by the different ocean biomes like mine is. You can see in my attached image how it is not exactly great to look at. If there is a way to filter out the ocean floor, then it should be added to the docs. If it is already there then I should learn how to read.
Selection_109

dgrimreaper:


overviewer_core/src/primitives/cave.c
```diff
@@ -78,15 +78,15 @@ cave_hidden(void* data, RenderState* state, int32_t x, int32_t y, int32_t z) {
      * at this point of the code the block has no skylight
      * but a deep sea can be completely dark
      */
-    if ((getArrayShort3D(state->blocks, x, y, z) == 9) ||
-        (get_data(state, BLOCKS, x, y + 1, z) == 9)) {
 
+    if ((getArrayShort3D(state->blocks, x, y, z) == 9) || (getArrayShort3D(state->blocks, x, y, z) == 8) ||
+        (get_data(state, BLOCKS, x, y + 1, z) == 9) || (get_data(state, BLOCKS, x, y + 1, z) == 8)) {
```
CounterPillow: Is the compiler smart enough to optimise away those repeated function calls?

overviewer_core/src/primitives/cave.c
```diff
@@ -78,15 +78,15 @@ cave_hidden(void* data, RenderState* state, int32_t x, int32_t y, int32_t z) {
...
         for (dy = y + 1; dy < (SECTIONS_PER_CHUNK - state->chunky) * 16; dy++) {
             /* go up and check for skylight */
             if (get_data(state, SKYLIGHT, x, dy, z) != 0) {
                 return true;
             }
-            if (get_data(state, BLOCKS, x, dy, z) != 9) {
+            if (get_data(state, BLOCKS, x, dy, z) != 8 && get_data(state, BLOCKS, x, dy, z) != 9) {
```
CounterPillow: Same here.

overviewer_core/src/primitives/cave.c
dgrimreaper: No idea, not a C programmer ;).  Want them in temporary variables?

tswsl1989: > Is the compiler smart enough to optimise away those repeated function calls?
Best answer I can give to that without checking a compiled version is "maybe".
I'd probably err on the side of a (clearly named) temporary variable either way.

dgrimreaper: ok will do

dgrimreaper:
Travis you suck.  Sigh pip fail.

Hopefully these temp var declarations are ok.

dgrimreaper:
All done

CounterPillow:
Merged, thanks for your contribution.

PR #1705, opened 2019-12-31
Track centers by layer instead of by world

cliffmeyers:
## Summary of issue

This fixes a bug where a world with multiple layers with different `northdirection` values would initially load at the wrong center point because only one center can be stored per world in the `overview.collections.centers` map.

See the config below. I've been assuming that multiple renders per world is intended usage.

```
renders["llc-latest-ul"] = {
    "world": "llc-latest",
    "title": "Latest UL",
    "northdirection": "upper-left",
    "defaultzoom": 7,
    "crop": (-500, -500, 500, 500),
    "center": [-250, 64, -250]
}

renders["llc-latest-lr"] = {
    "world": "llc-latest",
    "title": "Latest LR",
    "northdirection": "lower-right",
    "defaultzoom": 7,
    "crop": (-500, -500, 500, 500),
    "center": [-250, 64, -250]
}
```

When Overviewer loaded, it would display the "UL" map at an incorrect position `265/64/265`. This is because the center point in `overviewer.collections.centers["llc-latest"]` was the one calculated for the second layer which had a different orientation, hence flipping the coordinates around the origin.

See the code here that loops over the tile sets and eventually stores the centers.
https://github.com/overviewer/Minecraft-Overviewer/blob/751ba39bd06ed780c7746357f2a9b90ca032a549/overviewer_core/data/js_src/util.js#L357-L417

The `tileSets.forEach` executes for each layer and on line 412 the center is set per-world: the center point from the last layer defined for a world will always "win".

## The fix

The code changes are pretty simple: `overviewer.collections.centers` is now a map of maps, keyed by world name and layer name, so that each layer gets its own center point.

## Testing

I updated the Overviewer source and ran the following command to push the changes into the map I've been using for testing

`cat overviewer.js util.js > /mnt/c/dev/code/ov-test/output/overviewer.js`

The map defaults to the correct x/y now, and I can switch between layers and worlds and the coordinates load as expected. I haven't made use of overlays yet; not sure if this would have any impact to those.

### DIFF
--- a/overviewer_core/data/js_src/overviewer.js
+++ b/overviewer_core/data/js_src/overviewer.js
@@ -39,7 +39,7 @@ overviewer.collections = {
 
         /**
          * When switching regionsets, where should we zoom to?
-         * Defaults to spawn.  Stored as map of world names to [latlng, zoom]
+         * Defaults to spawn.  Stored as map of maps: world names to layer names to [latlng, zoom]
          */
         'centers': {},
 
--- a/overviewer_core/data/js_src/util.js
+++ b/overviewer_core/data/js_src/util.js
@@ -142,8 +142,11 @@ overviewer.util = {
 
 
                 // save current view for the current_world
-                overviewer.collections.centers[overviewer.current_world][0] = overviewer.map.getCenter();
-                overviewer.collections.centers[overviewer.current_world][1] = overviewer.map.getZoom();
+                let current_center = [overviewer.map.getCenter(), overviewer.map.getZoom()];
+                let current_layer = overviewer.current_layer[overviewer.current_world] ||
+                    Object.values(overviewer.collections.mapTypes[overviewer.current_world])[0];
+                let layer_name = current_layer.tileSetConfig.path;
+                overviewer.collections.centers[overviewer.current_world][layer_name] = current_center;
 
                 overviewer.layerCtrl.remove();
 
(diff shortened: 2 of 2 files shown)


CounterPillow:
Thanks for your PR. It took a while for me to review and merge because:

1. first there was uni stress
2. then my Lenovo ThinkPad T495 that I use for development broke one month into owning it because Lenovo builds cheap garbage so I had to send it in because yes I am going to make them regret saving those $0.20 on every component by using their warranty for years to come at the slightest failure of any component
3. then I decided to try and set up a dev environment on my Windows desktop instead of doing this in a VM which involved me having to hack Pillow because ????????????????????????????? loading watercolor.png fails on Windows mingw-w64 because a Pillow PNG DPI calculation somehow reads NaN and tries to `int()` that???????????
4. and now we're here, despite the continued terribleness of computers

cliffmeyers:
@CounterPillow no worries, sorry to hear about the PC troubles, and thanks for the merge! If there are any other client-side issues you wish could get fixed but don't have time to work on, please let me know and I'll do what I can.

FWIW, I've had some success using Windows Subsystem for Linux with Ubuntu to work on this and other projects.

CounterPillow:
If you're good at trigonometry, the worldfromlatlng and worldtolatlng (or whatever they were called) coordinate conversion functions need slight fixing; currently they're not actually the reverse of each other because as the code comment says, the problem was just plugged into wolfram alpha and it might not be the correct solution.

Otherwise the JS code needs a lot of general cleanup, since there's a lot of unused stuff and hacked together bits from the gmaps to leaflet move.

PR #1716, opened 2020-02-02
Fix chest rendering by using 'type' property

### LINKED ISSUE #1667: BUG: some chests not showing.
there are some places in our main base that do not show all the chests when I am positive the chests are there.
the missing chests should be in front of the signs.

image

Auron956:
This resolves a few issues with chest rendering that were introduced with Minecraft 1.13 granting the ability to place chests next to each other. Notably:
- Two single chests placed adjacently would be incorrectly rendered as one double chest
- Chest blocks with two or more adjacent chest blocks of the same type would fail to render at all

The solution here is to use the provided 'type' property of chests blocks to determine whether each is a single chest or the left/right half of a double chest, instead of the previous method which would check adjacent blocks for the presence of chest blocks.

I believe this also resolves #1667.


CounterPillow:
Thanks, confirmed to work and merged. As an aside note for anyone wondering, single chests next to each other can be placed by holding down shift while placing them. Had to look that one up.

I'm glad we can finally throw out a lot of the ancildata jank now.

PR #1717, opened 2020-02-03
Add Barrel Block

DragonDev1906:
- Barrel Block with 6 directions and open/closed state
- One of the missing Blocks in #1613


CounterPillow:
confirmed to be working and merged. I don't think the open/closed state actually matters since it appears to close as soon as a player leaves the inventory menu, like a chest would, but I guess it's nice that it's there.

PR #1747, opened 2020-04-05
Apply consisting sorting to markers JS files in genPOI

cliffmeyers:
genPOI: apply consistent sorting to markers.js and markersDB.js

use json.dumps(sort_keys=True) to apply a consistent sorting to the output JS files.
improves the readability of diffs and making config testing easier


cliffmeyers:
cc @CounterPillow

PR #1791, opened 2020-06-25
[WIP] Initial 1.16 block work

CounterPillow:
I've settled for using blockids 1000 and above for this.

- [x] Ancient debris
- [x] Basalt
- [x] Polished Basalt
- [x] Soul Campfire
- [x] Blackstone block
- [x] Netherite block

- [ ] the rest


georgekarrv:
can't wait to regen my nether with all these blocks rendered. Thanks for working on this!

CounterPillow:
Not everything has been added yet but I think we can probably merge this now, better than leaving it hanging here

PR #1803, opened 2020-07-04
Update building.rst

### LINKED ISSUE #1802: Build docs incorrect
The build document (<https://github.com/overviewer/Minecraft-Overviewer/blob/master/docs/building.rst>), states:

> If you choose to install pillow through pip instead of your distribution's package manager, you won't get the pillow headers...

I've tested building from source on several distro images (debian, ubuntu, fedora, centos and alpine) using the distro-supplied pillow and have found that it is necessary to copy three headers from <https://github.com/python-pillow> in every case. The docs should reflect this.

davemac30:
Fixes #1802

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -135,12 +135,17 @@ You will need the following packages on Debian-derived distributions (e.g. Ubunt
 * python3-numpy
 
 .. note::
-    If you choose to install pillow through pip instead of your distribution's package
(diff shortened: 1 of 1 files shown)


CounterPillow:
Please include that they should be of the precise Pillow version that was installed due to potential ABI breakage.

CounterPillow:
See discussion in issue

Issue #1665, opened 2019-10-28
Feature Request: Heat map layer of chunk change dates


bagobones:
Looking at the slim chunk layer I was wondering if it would be possible to generate a heat map of chunk updates based on last update.

IE chunks that have been updated in X days are read and older ones would be yellow and fade out.

I think it would be a neat layer to have to highlight on order servers where people are the busiest as players tend to spread out over time and it can be hard to find new builds etc.

If this was a feature it would ideally let you set a range of color for the heat map maybe in days and look something like the slime chunk layer.

enaut:
Sounds great! Do a fork and then a pull request ;-) I guess the maintainers will include it or help you get it to a state where it can be included.

Wunkolo:
This would be really cool actually. Some parts of my map are more "dead" than others and whenever I push a map update, people would have to swim around the map to see what changed. This would allow for people to get right to seeing all the new tiles that updated since the last render

bagobones:
@Tirpitz93  A history filter would require A LOT of tile data to be saved. but is also a cool idea.

What I am proposing is something that could run based off the existing delta scan, over-viewer should already be tracking when each chunk was updated somewhere. If I have time to look I assume the last update stamp shouldn't be much harder than finding the slime chunk info (unless that is just calculated on the seed).

I don't have a lot of time to look at code and python is more of a minor hobby for me.

Tirpitz93:
@bagobones I dont think that any change to the MC side of things would need to be carried out, just the web frontend, using something like rsnapshot or other rsync based tool. only new tiles are added each time.
I already use rsnapshot to offload the map to the render server before I run overviewer.
the only requirement would be to keep the old timestamped tiles to allow rollback.

Tirpitz93:
@bagobones I amgoing to open a feature request and fork see what I can come up with.

generrosity:
Each chunk had an idea of "inhabited time", which I think might be a viable indication of how busy a chunk had been. Might not be worth what was asked, but might be trivial to impliment

PR #1811, opened 2020-07-12
Add Heatmap RenderPrimitive (chunk mtime)

DragonDev1906:
As suggested in #1665 I've added a heatmap showing recent chunks.

- Based on the chunk mtime (instead of using the chunk inhabited time, which would have required calculating a min/max)
- Configured by setting an earliest timestamp (alpha=0) and a current timestamp (alpha=255)
- To make the implementation easier only the alpha value is changed with red as a base color (see Screenshot)
- You might have to use forcerender, otherwise the layer might not update correctly (maybe deleting the related files in the output might work, too)
- Could someone test this with a more active map? My testmap is rather small and had only 3 different chunk mtimes, which resulted in the sharp edges shown below. I would like to see how it looks on a more active map.

Example configuration:
```
renders["spawnable"] = {
   "world": "Test12985",
   "rendermode": [ClearBase(), HeatmapOverlay(
      # t_invisible=int((t_now - timedelta(days=30)).timestamp()),
      # t_full=int(t_now.timestamp()),
   )],
   "title": "Heatmap Overlay",
   "overlay": ["normalrender"],
}
```

### DIFF
--- a/overviewer_core/src/primitives/overlay-heatmap.c
+++ b/overviewer_core/src/primitives/overlay-heatmap.c
@@ -1,0 +1,93 @@
+/*
+ * This file is part of the Minecraft Overviewer.
+ *
+ * Minecraft Overviewer is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or (at
+ * your option) any later version.
+ *
+ * Minecraft Overviewer is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with the Overviewer.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <math.h>
+#include "overlay.h"
+
+typedef struct {
+    /* inherits from overlay */
+    RenderPrimitiveOverlay parent;
+    int t_invisible;
+    int delta_t;
+} RenderPrimitiveHeatmap;
+
+static void get_color(void* data, RenderState* state,
+                      uint8_t* r, uint8_t* g, uint8_t* b, uint8_t* a) {
+    RenderPrimitiveHeatmap* self = (RenderPrimitiveHeatmap*)data;
+    long mtime;
+    float _value_f;
+    char value;
+    PyObject *mtime_pyobj;
+
+    // Set default values (will get overridden based on self->mode)
+    *r = 255;
+    *g = 0;
+    *b = 0;
+
+    // Get the chunk modified time
+    mtime_pyobj = PyObject_CallMethod(state->regionset, "get_chunk_mtime", "ii", state->chunkx, state->chunkz);
+    if (mtime_pyobj == NULL) {
+        *a = 0;
+        return;
+    }
+    mtime = PyLong_AsLong(mtime_pyobj);
+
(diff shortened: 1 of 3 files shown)


overviewer_core/src/primitives/overlay-heatmap.c
```diff
@@ -0,0 +1,93 @@
...
+    PyObject *mtime_pyobj;
+
+    // Set default values (will get overridden based on self->mode)
+    *r = 255;
+    *g = 0;
+    *b = 0;
+
+    // Get the chunk modified time
+    mtime_pyobj = PyObject_CallMethod(state->regionset, "get_chunk_mtime", "ii", state->chunkx, state->chunkz);
+    if (mtime_pyobj == NULL) {
```
CounterPillow, line 43: Should check against `Py_None`

overviewer_core/src/primitives/overlay-heatmap.c
```diff
@@ -0,0 +1,93 @@
...
+        *a = 0;
+        return;
+    }
+    mtime = PyLong_AsLong(mtime_pyobj);
```
CounterPillow, line 47: You need to `PY_XDECREF(mtime_pyobj)` sometime after this, otherwise you leak a reference to `mtime_pyobj` I believe.

CounterPillow (changes requested):
Sorry it took a while to review this. I haven't personally run it yet, but there's two things I've spotted which are minor issues before I did. Thanks for your PR!

overviewer_core/src/primitives/overlay-heatmap.c
DragonDev1906: Thanks for the Info, yes that should indeed check for `Py_None` according to the docs  `PyObject_CallMethod`returns NULL on errors or exceptions. Since it probably doesn't hurt I have changed it to check for `NULL` and `PY_None`, catching exceptions in python should any occur and checking for None as return value.

overviewer_core/src/primitives/overlay-heatmap.c
DragonDev1906: Thanks, I believe that is correct

DragonDev1906:
I applied the suggested changes and rebased the PR to the current version (to test if it still works)

overviewer_core/src/primitives/overlay-heatmap.c
CounterPillow: I highly suspect Py_None is simply an alias to NULL, I think checking for NULL is fine, in retrospect.

CounterPillow:
Thanks. I'll try to render a map with it today to see it in action, and then merge it if it's all good

CounterPillow:
```
overviewer_core/src/primitives/overlay-heatmap.c: In function ‘get_color’:
overviewer_core/src/primitives/overlay-heatmap.c:48:5: warning: implicit declaration of function ‘PY_XDECREF’; did you mean ‘Py_XDECREF’? [-Wimplicit-function-declaration]
     PY_XDECREF(mtime_pyobj);
     ^~~~~~~~~~
     Py_XDECREF
```
Small mistake, please change and git commit -a --amend.

Trying this out on a map now anyway with it fixed locally.

DragonDev1906:
Thank you for pointing this out, apparently I forgot to run `python setup.py build`

CounterPillow:
image
Seems to work fine, aside from the fact that Minecraft is so write-happy you rarely get smooth transitions since essentially every chunk a player can see will be written.

DragonDev1906:
Minecraft probably has to save whenever there is any entity in the chunk (or might just save always to make programming easier).

CounterPillow:
It famously updates the chunk whenever a cow moos, from what I've heard

Issue #1825, opened 2020-08-06
Stairs not rendering correctly


matej-darebnik:
Hi, after merge from 116-blocks yesterday, stairs stopped rendering correctly:
![image](https://user-images.githubusercontent.com/51075039/89570999-3dc70a00-d827-11ea-805b-30e0ad43523a.png)

This is how it looks ingame:
![image](https://user-images.githubusercontent.com/51075039/89570952-2c7dfd80-d827-11ea-9767-b68f0022c681.png)

Version of overviewer:
Minecraft Overviewer 0.16.0 (50b84e9)
built on Wed Aug  5 17:15:03 2020

matej-darebnik:
I have just noticed that my wooden roofs do not seem to be affected, so I have tried rendering dark oak stairs and also tried cobblestone. It seems that some types of stairs render correctly, while others look weird...

![image](https://user-images.githubusercontent.com/51075039/89572843-fd1cc000-d829-11ea-9f4b-9ec068f2f535.png)

![image](https://user-images.githubusercontent.com/51075039/89572540-854e9580-d829-11ea-9272-02e329b4e000.png)

IncredibleHolg:
I will try to pin down the commit, that introduced this bug.

PR #1826, opened 2020-08-06
Resolve stair & stem rendering issues

Auron956:
This resolves a couple of issues with some of the recent 1.16 commits:
- Some stairs were incorrectly being identified as walls, causing incorrect pseudo data to be assigned & resulting in rendering issues
- A check for warped & crimson stem blocks was too inclusive, resulting in stems for melons, pumpkins, and mushrooms to no longer render at all

Fixes: #1825

### DIFF
--- a/overviewer_core/src/block_class.c
+++ b/overviewer_core/src/block_class.c
@@ -62,7 +62,7 @@ bool block_class_is_wall(mc_block_t block) {
     mc_block_t mask = 0b11111111;
     mc_block_t prefix = 0b111 << 8;     // 1792 is the starting offset
     // if the xor zeroes all bits, the prefix must've matched.
-    return (block & ~mask) ^ prefix == 0;
+    return ((block & ~mask) ^ prefix) == 0;
 }
 
 const mc_block_t block_class_stair[] = {
--- a/overviewer_core/src/overviewer.h
+++ b/overviewer_core/src/overviewer.h
@@ -31,7 +31,7 @@
 
 // increment this value if you've made a change to the c extension
 // and want to force users to rebuild
-#define OVERVIEWER_EXTENSION_VERSION 90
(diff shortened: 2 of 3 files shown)


IncredibleHolg:
@Auron956 out curiosity: do you hava a sharp eye or did you use "git bisect" for the stair bug?

Auron956:
@IncredibleHolg I hadn't used `git bisect` for this (in truth I wasn't aware of it, so good to know!), instead assuming that one of the commits in the 116-blocks branch was responsible meant there were reasonably few commits to go through and even fewer that touched anything remotely related to stairs.

It helped to know that stairs were one of a handful of blocks that rely on 'pseudo data' generated within `iterate.c` to render correctly. Seeing that the only change here was the introduction of `block_class_is_wall()` meant that it was possibly related to the issue, which testing later confirmed.

wolfen351:
While you are in this headspace, could you add on the tops of the bamboo sticks?

LasseRheinstroem:
I think there are more flawed.
Polished Diorite Stairs are OK now, but the simple Diorite Stairs are still wrong

Auron956:
> simple Diorite Stairs are still wrong

Looks fine to me when using 0.16.3, are you sure the area of your map containing these stairs has been re-rendered with the latest version? If so, are you able to share which arrangement of stair blocks you're using that still looks incorrect?

LasseRheinstroem:
i can confirm the latest version (i got the "2020-08-08 12:36:16  Welcome to Minecraft Overviewer version 0.16.3 (4c0a69d)!" )
the polished roofs were "updated" when rendering, but the simple ones are probably not.

>  are you able to share which arrangement of stair blocks you're using that still looks incorrect?

it's just the roof of a house

I have now started a new full render, let's see what comes out of it

PR #1835, opened 2020-08-15
More nether

IncredibleHolg:
add advanced blocks, nether roof update
    
* crying_obsidian, lodestone, respawn_anchor
* ancient_debris part of the nether roof
Includes PR #1834 and #1832

Issue #1807, opened 2020-07-11
add 1.16 blocks to main branch


IncredibleHolg:
@CounterPillow I am opening this isse to add the missing blocks to the main branch. The code discussed in #1772 looks not yet stable, so adding the missing blocks in the ``classical`` way may get usable results faster than incorporating the modernGLRender branch.

Would you like to squash/merge up to a3e4812ca906146408171b34e3c313fcb5e7ee69  into the master branch before moving on?

Missing blocks:
* Nylium, Hyphae, shroomlight  and wart blocks
* All new logs
* All new wood types, new craftable wood blocks
* New craftable stone types

CounterPillow:
Please just send a PR to the 116-blocks branch.

IncredibleHolg:
Starting with the easy ones, warped and crimson nylium
![nylium-01](https://user-images.githubusercontent.com/41533138/87243192-38e38600-c434-11ea-8cb5-157392b6a048.png)
![nylium-02](https://user-images.githubusercontent.com/41533138/87243195-3da83a00-c434-11ea-9cab-06a4462b7c6d.png)

IncredibleHolg:
Added logs aka stems
![nether_stems01](https://user-images.githubusercontent.com/41533138/87248535-34ca5f00-c45a-11ea-8262-f0625c87ae19.png)

![nether_stems02](https://user-images.githubusercontent.com/41533138/87248541-3ac04000-c45a-11ea-8dc7-ee79705d1854.png)

IncredibleHolg:
Added hyphae
![nether_stems03](https://user-images.githubusercontent.com/41533138/87249553-05b6ec00-c460-11ea-9419-69437a210e82.png)

IncredibleHolg:
crimson and warped biomes blocks are implemented PR #1812
![netherbiomes01](https://user-images.githubusercontent.com/41533138/87729004-9a08b200-c7c4-11ea-8fd7-595f27052a25.png)

LeadMagnet:
I'm having issue with Polished Blackstone Bricks.

[What it looks like in game](https://photos.app.goo.gl/to9b5JtgjXD8iCT26)

[What it looks like after Overviewer render](http://map.texter.family/worlds/#/-85/64/-211/max/world%20-%20overworld/world)

Running version 0.15.88 on W2019 server.

IncredibleHolg:
No surprise, they are not yet implemented

LeadMagnet:
[...]

PR #1839, opened 2020-08-23
Soul lightning

IncredibleHolg:
Add soul_fire, soul_lanterns and soul_torch

included PRs: #1835 and #1832 , I close this. 

Related Issue: #1807


overviewer_core/src/mc_id.h
```diff
@@ -360,12 +396,16 @@ enum mc_block_id {
     block_jungle_sign = 11404,
     block_acacia_sign = 11405,
     block_dark_oak_sign = 11406,
+    block_crimson_sign = 12505,
+    block_warped_sign = 12506,
```
CounterPillow, line 400: Any particular reason for using high IDs for these and the wall signs? In general we should avoid high IDs because we create an array up to the highest block index, which increases memory usage if this is a new highest ID block.

overviewer_core/textures.py
```diff
@@ -1994,12 +2020,18 @@ def composter(self, blockid, data):
...
-    side2 = self.transform_image_side(firetextures[1]).transpose(Image.FLIP_LEFT_RIGHT)
+    if blockid == 51:
+        firetextures = self.load_fire()
+        side1 = self.transform_image_side(firetextures[0])
+        side2 = self.transform_image_side(firetextures[1]).transpose(Image.FLIP_LEFT_RIGHT)
+    elif blockid == 1040:
+        soul_firetextures = self.load_soul_fire()
+        side1 = self.transform_image_side(soul_firetextures[0])
+        side2 = self.transform_image_side(soul_firetextures[1]).transpose(Image.FLIP_LEFT_RIGHT)
+
```
CounterPillow, line 2034: I don't think you need to duplicate the transform code here.

```python
firetextures = self.load_fire() if blockid == 51 else self.load_soul_fire()
side1 = self.transform_image_side(firetextures[0])
side2 = self.transform_image_side(firetextures[1]).transpose(Image.FLIP_LEFT_RIGHT)
```
should be enough to handle both cases.

overviewer_core/textures.py
```diff
@@ -2048,6 +2082,8 @@ def stairs(self, blockid, data):
         164: "assets/minecraft/textures/block/dark_oak_planks.png",
         180: "assets/minecraft/textures/block/red_sandstone.png",
         203: "assets/minecraft/textures/block/purpur_block.png",
+        509: "assets/minecraft/textures/block/crimson_planks.png",
+        510: "assets/minecraft/textures/block/warped_planks.png",
```
CounterPillow, line 2086: why are these at 500 as opposed to 1000? not that it really matters, just wondering if there's something you were planning on doing with that

CounterPillow (commented):
Thanks for your work on this, I'd like to know more about some of the ID choices you've made though (see comments left on code)

overviewer_core/textures.py
LieuweWestra: These are the block IDs that minecraft uses itself. Only in bedrock these days, but I thought it was best to just continue the existing pattern.

overviewer_core/textures.py
IncredibleHolg: Hi @CounterPillow 

Unless i completely missed something about the ``data`` parameter, this is always zero for both fire variants ( not 0 .. 15)
So this can be simplified to

```python
# fire and soul_fire
@material(blockid=[51, 1040], transparent=True)
def fire(self, blockid, data):
    if blockid == 51:
        textureNS = self.load_image_texture("assets/minecraft/textures/block/fire_0.png")
        textureEW = self.load_image_texture("assets/minecraft/textures/block/fire_1.png")
    elif blockid == 1040:
        textureNS = self.load_image_texture("assets/minecraft/textures/block/soul_fire_0.png")
        textureEW = self.load_image_texture("assets/minecraft/textures/block/soul_fire_1.png")
    side1 = self.transform_image_side(textureNS)
    side2 = self.transform_image_side(textureEW).transpose(Image.FLIP_LEFT_RIGHT)
    img = Image.new("RGBA", (24,24), self.bgcolor)
    alpha_over(img, side1, (12,0), side1)
    alpha_over(img, side2, (0,0), side2)
    alpha_over(img, side1, (0,6), side1)
    alpha_over(img, side2, (12,6), side2)    
    return img
``` 

Or am I missing something?

IncredibleHolg:
Added the changes for soul fire texture loading 
sha1 d23fdb1

PR #1842, opened 2020-08-26
web_assets: update leaflet to 1.6.0

ArchmageInc:
Update LeafletJS from 1.5.1 to 1.6.0

Minor bug fixes and improvements, no breaking changes.


CounterPillow:
Thanks for the pull request, I've updated Leaflet to 1.7.1 in master just now, so this is no longer required. It did remind me though to check for a new release!

PR #1846, opened 2020-09-02
Remaining 1.16 blocks (from Auron956)

Esvandiary:
This is almost entirely taken from the [`blocks-116` branch](https://github.com/Auron956/Minecraft-Overviewer/tree/blocks-116) by @Auron956.

Changes from that branch:
- Adds block IDs for remaining 1.16 blocks
- _Lots_ of changes to `textures.py`, including some 1.16 stuff but also reworking some existing logic (without changing behaviour) and formatting changes
- A few tweaks to `world.py` along similar lines

I also added one commit on the end which adds Ancient Debris to the nether roof list (it can occasionally generate up to Y=~120 which currently results in mid-air roof pillars on Nether renders).

I can try to extract the 1.16-relevant parts and skip the formatting/logic tweaks if that's preferred. IMO the formatting changes are mostly for the better, but it sure is a lot of noise.


Auron956:
Not what I expected to see but okay. For what it's worth I abandoned this branch when I realized @IncredibleHolg had already been working on implementing 1.16 blocks - I didn't feel it made sense to have multiple PRs for the same block additions but do with the branch as you wish.

Should also note that it's incomplete, from memory chains & various potted plants introduced in 1.16 were never implemented in this branch.

Esvandiary:
> Should also note that it's incomplete, from memory chains & various potted plants introduced in 1.16 were never implemented in this branch.

Ah, thanks for the info @Auron956 - I wasn't aware anyone else was working on 1.16 blocks. I saw your branch on the GH Network page and failed to spot anything else that looked as complete as yours. I'll close this for now to let @IncredibleHolg finish their work, and have another look at how things are going in a few weeks.

Sorry for the confusion!

Issue #1874, opened 2020-11-23
Crash right at the end of the process


wolfen351:
2020-11-23 21:40:17  Looking for entities in <overviewer_core.world.CroppedRegionSet object at 0x7f3f52012be0>...
2020-11-23 21:40:18  Buckets has 97158 entries.
2020-11-23 21:40:18  Buckets has 97319 entries.
2020-11-23 21:40:18  Buckets has 96173 entries.
2020-11-23 21:43:55  All the threads completed.
2020-11-23 21:43:55  Done.
2020-11-23 21:43:55  Initialized an empty UUID cache.
*******************************************************************************
2020-11-23 21:44:03 E An error has occurred. This may be a bug. Please let us know!
See http://docs.overviewer.org/en/latest/index.html#help

This is the error that occurred:
Traceback (most recent call last):
  File "/MOV/overviewer_core/nbt.py", line 199, in read_all
    tagtype = ord(self._file.read(1))
TypeError: ord() expected a character, but string of length 0 found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./overviewer.py", line 653, in <module>
    ret = main()
  File "./overviewer.py", line 168, in main
    g.genPOI.main()
  File "/MOV/overviewer_core/aux_files/genPOI.py", line 599, in main
    handlePlayers(worldpath, list(worldpath_filters), markers)
  File "/MOV/overviewer_core/aux_files/genPOI.py", line 325, in handlePlayers
    data = PlayerDict(nbt.load(os.path.join(playerdir, playerfile))[1])
  File "/MOV/overviewer_core/nbt.py", line 31, in wrapper
    return func(fileobj, *args)
  File "/MOV/overviewer_core/nbt.py", line 40, in load
    return NBTFileReader(fileobj).read_all()
  File "/MOV/overviewer_core/nbt.py", line 207, in read_all
    raise CorruptNBTError("could not parse nbt: %s" % (str(e),))
overviewer_core.nbt.CorruptNBTError: could not parse nbt: ord() expected a character, but string of length 0 found
*******************************************************************************

LeadMagnet:
Can you upload the config file you're using?

wolfen351:
Here is the config - but I don't think it's a config error - I think its an empty player.dat file:

```import os

worlds["world"] = "/data/world"
outputdir = "/mc-map-data"

########### POI ###################################

def playerIcons(poi):
        if poi['id'] == 'Player':
[...]

PR #1876, opened 2020-11-25
Handle empty/corrupt player files better

wolfen351:
Handle bad files better - this is against Issue #1874


CounterPillow:
Not like this please - bare except is super bad in python because it catches more than just exceptions, e.g. KeyboardInterrupts.

CounterPillow:
I just added `nbt.CorruptNBTError` to the list of caught exceptions. The correct way to catch all exceptions (which I don't think we should do, if there's a case we can handle better we should handle it better) is to do `except Exception:` instead of bare except.

PR #1880, opened 2020-12-03
Fix typo in documentation

falsovsky:
In the **Introduction** section of the first page of the documentation there's a typo.
> For a simple example of what your renders will look like, head over to The **“Exmaple”** Map
http://docs.overviewer.org/en/latest/

This PR corrects it.


CounterPillow:
Deliberate pun

PR #1890, opened 2021-01-15
Minor spelling fixes

lenzls:
Fixes two minor spelling mistakes in the `index.rst` of the documentation.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -53,7 +53,7 @@ Documentation Contents
 Features
 ========
 
-* Renders high resolution images of your world, let's you "deep zoom" and see
+* Renders high resolution images of your world, lets you "deep zoom" and see
   details!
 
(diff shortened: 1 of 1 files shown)


CounterPillow:
I'm glad you realised that exmaple is deliberate before anyone told you. :)

Though, merging two commits for one grammatical fix and a pointless capitalisation (that I disagree with) is not really something I want to do, so if you can force-push your branch with just a fix that's for "let's" then I'd much prefer that.

lenzls:
Ha, there is even a label for it!

I force pushed the branch so that only the real spelling issue is touched.

Let me know if you have further suggestions.

Btw: If you disagree with the capitalization in "ExMaple", you might also want to have a look at the capitalization in https://github.com/overviewer/Minecraft-Overviewer/wiki/Map-examples#official-overviewer-example-map-exmaple.

PR #1891, opened 2021-01-16
Added all new 1.16 wood blocks

QuentiumYT:
I've added 1.16 blocks using crimson and warped wood

This includes planks, slabs, stairs, plates, buttons, fences and gates

- [x] Tested compilation
- [X] Ran on a world with all blocks

(Ran on Debian Buster 10 with 1.16.4 world and 1.16.4 textures)

### DIFF
--- a/overviewer_core/src/mc_id.h
+++ b/overviewer_core/src/mc_id.h
@@ -281,6 +281,8 @@ enum mc_block_id {
     block_crimson_roots = 1019,
     block_soul_soil = 1020,
     block_nether_gold_ore = 1021,
+    block_crimson_slab = 1022,
+    block_warped_slab = 1023,
     // adding a gap in the numbering of walls to keep them all
     // in one numbering block starting at 1792
     // all blocks between 1792 and 2047 are considered walls
@@ -377,10 +379,7 @@ enum mc_block_id {
     block_honey_block = 11504,
     block_sweet_berry_bush = 11505,
     block_campfire = 11506,
-    block_bell = 11507,
-    // 1.16 blocks below
-    block_crimson_slab = 11508,
-    block_warped_slab = 11509
+    block_bell = 11507
 };
 
 typedef uint16_t mc_block_t;
--- a/overviewer_core/src/mc_id.h
+++ b/overviewer_core/src/mc_id.h
@@ -377,7 +377,10 @@ enum mc_block_id {
     block_honey_block = 11504,
     block_sweet_berry_bush = 11505,
     block_campfire = 11506,
-    block_bell = 11507
+    block_bell = 11507,
+    // 1.16 blocks below
+    block_crimson_slab = 11508,
+    block_warped_slab = 11509
 };
 
 typedef uint16_t mc_block_t;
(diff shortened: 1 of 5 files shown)


overviewer_core/src/mc_id.h
```diff
@@ -377,7 +377,10 @@ enum mc_block_id {
     block_honey_block = 11504,
     block_sweet_berry_bush = 11505,
     block_campfire = 11506,
-    block_bell = 11507
+    block_bell = 11507,
+    // 1.16 blocks below
+    block_crimson_slab = 11508,
+    block_warped_slab = 11509
```
CounterPillow, line 383: Can you use lower IDs? We currently store block sprites in an array, so if we keep making them higher we have a lot of unused space in it. Thanks!

CounterPillow (changes requested):
Let's try to keep the IDs small, that way we don't balloon the blocks array even further. Thanks for the PR though, and sorry it took me this long to get around to it!

overviewer_core/src/mc_id.h
QuentiumYT: Hello, I just increased the IDs as it was done before, do you mean filling ids from 1806 to 11336?

CounterPillow: Use 1000+, I added some 1.16 blocks there before. Don't use IDs between 1792 and 2047, those are all considered walls due to the wall checking function using a bit operation.

QuentiumYT:
I lowered down these 2 slabs, I see #1901's got a lot of 1.16 blocks as well (more than me)
If they fix the conflict, the PR could be richer in additions

overviewer_core/textures.py
```diff
@@ -2024,7 +2033,7 @@ def fire(self, blockid, data):
 # polished_granite polished_andesite polished_diorite granite diorite andesite end_stone_bricks red_nether_brick stairs
 # smooth_red_sandstone_stairs
 @material(blockid=[53, 67, 108, 109, 114, 128, 134, 135, 136, 156, 163, 164, 180, 203, 11337, 11338, 11339,
-          11370, 11371, 11374, 11375, 11376, 11377, 11378, 11379, 11380, 11381, 11382, 11383, 11384, 11415], 
+          11370, 11371, 11374, 11375, 11376, 11377, 11378, 11379, 11380, 11381, 11382, 11383, 11384, 11415, 11518, 11519],
```
CounterPillow, line 2036: you still use the old ones here and in other places

overviewer_core/world.py
```diff
@@ -853,6 +857,17 @@ def __init__(self, regiondir, rel):
...
+            'minecraft:crimson_pressure_plate': (11510, 0),
+            'minecraft:warped_pressure_plate': (11511, 0),
+            'minecraft:crimson_button': (11512, 0),
+            'minecraft:warped_button': (11513, 0),
+            'minecraft:crimson_fence': (11514, 0),
+            'minecraft:warped_fence': (11515, 0),
+            'minecraft:crimson_fence_gate': (11516, 0),
+            'minecraft:warped_fence_gate': (11517, 0),
+            'minecraft:crimson_stairs': (11518, 0),
+            'minecraft:warped_stairs': (11519, 0)
```
CounterPillow, line 870: move these to the lower IDs as well, thanks

CounterPillow:
I think these are in the master branch now courtesy of @IncredibleHolg, thank you for your contribution anyway, I'll refer back to it should there be any rendering issues in the ones that have been merged, and it finally pushed me to look into getting some stuff merged again.

QuentiumYT:
I'm sorry for the delay, got a few problems. I saw IncredibleHolg PR a few days after my own implementation and it seems to work after testing. If you still need my code, I can still work on it ^^
Nice to hear that :)

Issue #1676, opened 2019-11-12
Overlapping tiles all over the rendered map


hieutvan:
Hi there ;)

I've used Overviewer for quite a while and I found it's great! However, the rendered map usually has corrupt and overlapping tiles (as shown in the screenshot below). I've tried force-render, as well as deleting the render folder. I always execute the rendering process at midnight, at which no players are present. Please help me soon! Thanks!

I'm using MC 1.14.4, OV is built from source

![1226869725](https://user-images.githubusercontent.com/45692885/68640623-f717de00-053a-11ea-8549-843a915ceace.png)

Tantriss:
Hi, do you render directly from the world folder or do you copy the world folder in another location and render from here ?
Because I had the same problem until I updated my script to `rsync` the world folder and render from the copy.

hieutvan:
Wow so OV has to render from a static folder? That worked! Thanks a lot!

I think this probably should be added into the Readme.

Doregon:
@hieu-van A little late but I have just added a note and I hope that my pull request will be accepted. The same happened to me.

PR #1900, opened 2021-02-05
A little note on the Usage page

Doregon:
I noticed when I was rendering with my Minecraft server online, certain chunks would fly around the map and sit in the wrong location. I do not have a screenshot but this is just like issue #1676, and the suggestion made to add a note was not accompanied by a pull request or commit with one. 

This pull request adds a note on "docs/using.rst" saying that:

* The world should not be in use.
* Any programs that are using it should be closed.
* If the world is not able to be closed (like with a server), copy the world to a new location.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -42,6 +42,11 @@ you want to render).
 
     overviewer.py /home/username/mcserver /home/username/mcmap
 
+A little note: If the world folder is in use by another program at the time of
+rendering, you'll see some unexpected results with chunks being in the wrong
+places. Before you begin, make sure that you aren't using the world, or copy
+the world to a different location.
+
 After you enter one of the commands, The Overviewer should start rendering your
 map. When the render is done, open up *index.html* using your web-browser of
 choice.  Pretty cool, huh? You can even upload this map to a web server to share
--- a/overviewer_core/src/block_class.c
+++ b/overviewer_core/src/block_class.c
@@ -74,6 +74,8 @@ const mc_block_t block_class_stair[] = {
     block_spruce_stairs,
     block_birch_stairs,
     block_jungle_stairs,
+    block_crimson_stairs,
+    block_warped_stairs,
     block_quartz_stairs,
     block_acacia_stairs,
     block_dark_oak_stairs,
@@ -96,7 +98,10 @@ const mc_block_t block_class_stair[] = {
     block_end_stone_brick_stairs,
     block_red_nether_brick_stairs,
     block_mossy_stone_brick_stairs,
-    block_smooth_sandstone_stairs};
+    block_smooth_sandstone_stairs,
+    block_blackstone_stairs,
+    block_polished_blackstone_stairs,
+    block_polished_blackstone_brick_stairs};
 const size_t block_class_stair_len = COUNT_OF(block_class_stair);
 
 const mc_block_t block_class_door[] = {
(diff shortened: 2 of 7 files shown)


Doregon:
And nice PR number, we're finally in the 20th century my peeps.

Doregon:
Closing this pull request in favor of a new one, containing other changes.

CounterPillow:
> The world should not be in use.
> Any programs that are using it should be closed.
> If the world is not able to be closed (like with a server), copy the world to a new location.

Literally none of these points are true though.

Doregon:
Then how come with my Minecraft world open in single-player or when my Minecraft server is online, chunks are overlapping and in the wrong location? The fix was to do what I've described. If you don't believe me, try it yourself.

CounterPillow:
> Then how come with my Minecraft world open in single-player or when my Minecraft server is online, chunks are overlapping and in the wrong location? The fix was to do what I've described. If you don't believe me, try it yourself.

I have not analysed the problem on a deep enough level yet to definitively put my finger on what's going on, but I know enough to shoot down world copying cargo-culting suggestions:
- Minecraft regions store chunk locations in a small header at the start of the file that is not part of the NBT
- if Overviewer (or any application, for that matter, e.g. a file copy) were to read a chunk while it is being written, it might see an inconsistent view of essentially two versions of the file. However, this is pure speculation at this point; it's possible something entirely different is happening.
- Users have reported that copying the world and then rendering from the offline world still makes the chunk shuffling happen, which means either the problem is not related to what I think I'm happening or even a file copy in this case would create an inconsistent view of the world
- Reliable information is hard to come by because everyone just cargo cults "this worked for me" "solutions" when in reality the reason they worked might as well be that they got lucky on a copy or render.

There are a few things I want to try like making Overviewer load the entire region file into memory as fast as possible to hopefully mitigate some of this. I would however strongly recommend against making any definitive statements regarding Overviewer and rendering online worlds; telling users to go through multiple ugly steps to render a world should only be done once we actually definitively understand the problem and know this actually helps for sure, which at the moment is not the case even if we base our evidence on anecdotes from users.

Doregon:
Then maybe I'm wrong on the copying part and the save-off/render/save-on trick should be put in there instead. Either way there should be something saying "This might happen, this is a solution." 

And I understand if you don't have the time, and it might be a good idea to let some other people accept pull requests and help moderate the code with you, and then even when you're swamped with other things to do, someone can answer pull requests and compare code-- @dreamwraith said that they were up to help a little and I am too. Especially since (again, not your fault because you're busy) everyone has to go through so many hoops to find @IncredibleHolg's code if they haven't been here before and want to be able to use the 1.16 blocks, it would be easier to see multiple people reviewing and something happening.

And I'm also sorry if my copy-the-world idea came off as a cargo-culting solution. Maybe if I were to spend more time trying to figure out the problem, I could come with a pull request to fix it and allow users to just run the render, no questions asked.

I'm simply just trying to help, not trying to seem like a douche or begging to be added to the contributors list, I just came across a problem, found a few answers that said "copying is good" and ran tests to see if it was.

dreamwraith:
@Doregon - regarding the subject at hand (corrupt chunks) afaik it has to do with the world being SAVED while generating the map based on discussion in one of the other posted issues.

Once I added "disable saving" and "enable saving" to my pre and post run hooks, the corrupt chunk issue went away.

@CounterPillow - Hope you are doing well, good to see you :)

Doregon:
Yeah, I'll tweak my docs changes to say the same. Thanks, @CounterPillow @dreamwraith

Doregon:
Then it's not technically wrong that I said "close the world down". I'll tweak my changes to say use the save-off/save-on trick but I can still say "don't be using the world" or even elaborate on the subject.

PR #1941, opened 2021-06-06
Rworked glass pane connection

buscher:
- use in game data instead of manually calculating the adjacency data

This makes the code lighter and also fixes some connections for new blocks.

For example,

In-Game:
glass_panes_reworked_ingame

Without the rework (200% zoom)
glass_panes_reworked

With the rework (200% zoom)
glass_panes_reworked_fixed


CounterPillow:
Thanks, tested and code looks good, merged.

PR #1959, opened 2021-07-15
Fixed edge-lines and side rendering for snow layers

buscher:
Fix "dirty" snow, because of wrong edge-lines and missing texture side darkening.

Pretty hard so see in game, but for comparison, In-Game:
dirty_snow_ingame

Current master (without the fix):
dirty_snow

With the fix:
dirty_snow_fixed


generrosity:
Applogies for the query, but wondering @buscher if part of the the 'edges' you fixing was these other visual gaps and alignments? Althought at a tight zoom of your fix, I suspect it is unrelated and I can log a new Issue for it

image
image

buscher:
I can see the issues you are talking about. But yes, I did not touch this part of the code.
I will have to touch this code part again after adding powder_snow, then maybe I will look deeper into it and see if I can fix it, but this will have to come laaaaaaater, got lots of other stuff on my TODO list first :)

PR #1966, opened 2021-07-29
fix marker center

ProkopRandacek:
For some reason, the margin -50% doesn't work and the markers are offset just by something like -10%. I have no idea why but switching to translate fixes the issue.


ProkopRandacek:
Bump.

I checked how dynmap does this and they use the same exact solution as I am proposing.

PR #1967, opened 2021-07-29
fix warning foreground color

ProkopRandacek:
The logging colors should explicitly set the foreground colors. Otherwise, terminals that have bright foreground color by default end up writing white text to yellow background and I cant read it.

before:

after:

### DIFF
--- a/overviewer_core/logger.py
+++ b/overviewer_core/logger.py
@@ -44,9 +44,9 @@ COLORIZE = {
     'DEBUG': CYAN,
 }
 HIGHLIGHT = {
-    'CRITICAL': RED,
-    'ERROR': RED,
-    'WARNING': YELLOW,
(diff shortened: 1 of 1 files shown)


CounterPillow:
This strictly makes warnings worse for me
image

ProkopRandacek:
I don't think there is a way of getting the color theme rgbs from the host terminal (which could be used to determine if black or white text should be used).
But I believe that this dark orange is not very common.

Maybe use dark text by default and add a `--white-warning` flag to change it? Or the other way around.

CounterPillow:
My linux tty is also using orange for this.

I'd rather not introduce yet another option for the sake of colors

ProkopRandacek:
Yes my tty is also orange.

I don't think there is a good solution to this then.

PR #1972, opened 2021-08-15
Update the default color list used for the Biome Overlay.

poempelfox:
Instead of the old, only partially complete and illogical default scheme, we now use the same one as Amidst.

The old default color scheme only contained a fraction of the available biomes, it seems it hasn't been updated for newer minecraft versions in ages. At first I tried to fill in the gaps by selecting colors myself to match the existing ones, but I was soon pulling my hair out, because there seemed to be no logic in these color choices. So instead, I threw it all away and just used the color scheme from the Amidst project that they describe in https://github.com/toolbox4minecraft/amidst/wiki/Biome-Color-Table . That scheme does seem to have a certain logic (e.g. the colors of similar biomes are similar), and it also seems to be sort of a de facto standard - e.g. the chunkbase seedmap uses it too.

While no code was directly copied, so IMHO their GPL3 license does not apply, I only found it fair to give them credit - which I did in both the source code and the documentation.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1244,8 +1244,10 @@ BiomeOverlay
 
     biomes
         A list of ("biome name", (r, g, b)) tuples to use as colors. Any
-        biome not specified won't be highlighted. If not provided then 
-        a default list of biomes and colors is used.
+        biome not specified won't be highlighted. If no list is provided,
+        then a default list of biomes and colors is used, that is trying
+        to match the de-facto-standard for biome coloring established
+        by the Amidst project: https://github.com/toolbox4minecraft/amidst/wiki/Biome-Color-Table
 
         Example::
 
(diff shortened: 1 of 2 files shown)


fireduck64:
Does this fix the biome color issue here? https://github.com/overviewer/Minecraft-Overviewer/issues/2022

I am just a user and noticed some work in that ticket and this PR want am worried they are not working together.

poempelfox:
> Does this fix the biome color issue here?

It most certainly does not, because it hasn't got the slightest thing to do with biome tinting, and is from MC 1.17 times.

The biome overlay is affected by the same problem with MC 1.18 though, and will not recognize any biome in files written by MC 1.18 because the file format seems to have changed.
Even after the general problems with MC 1.18 files have been fixed and this PR has been merged, there will need to be another update to support MC 1.18 for biome overlays, because that version has essentially renumbered most biomes.

Issue #2011, opened 2021-12-15
Empty tiles with flat world


Doelia:
I created a world with Minecraft 1.8.1 vanilla : "Solo / Create new world / More world Options... / World Type : Superflat", defaults settings.

Overviewer generate empty tiles.

<img width="867" alt="Screenshot 2021-12-15 at 19 23 54" src="https://user-images.githubusercontent.com/4354254/146243643-09374513-8024-4535-b6ce-1e8116e8cb3d.png">

It works with a normal world.
I try with Windows binaries and compiling from source on Debian.

No error or warning messages :
```
overviewer.py "world" ./render
2021-12-15 18:25:23  Welcome to Minecraft Overviewer version 0.17.40 (f6ee2cf)!
2021-12-15 18:25:23  Generating textures...
2021-12-15 18:25:42  Preprocessing...
2021-12-15 18:26:01  Rendering 581 total tiles.
 99% [======================================================================================================= ] 577 11.20t/s eta 00h 00m 00s
2021-12-15 18:26:53  Rendering complete!
2021-12-15 18:26:54  Your render has been written to '/render', open index.html to view it.
```

spacegriffin:
Do superflat worlds in 1.18 generate below y=0? If so, that's the problem. Overviewer doesn't currently render blocks below y=0 or above y=256

greatmastermario:
Yes, they generate from Y=-64 as of 1.18. This is fixed in my PR (see screenshots) https://github.com/overviewer/Minecraft-Overviewer/pull/2012

lostallmymoney:
Normal world and I get the same result on 1.18.1

Issue #2003, opened 2021-12-07
Blocks above y=254 do not render


jack-webb:
I think related to #1954, blocks at or above 255 aren't rendering on my 1.18 world. I'm using master a5b36f5. This also doesn't render below y=0

| Render | In-game |
| - | - | 
| ![UUIsA](https://user-images.githubusercontent.com/23270214/145108403-ba7b0e1d-5f31-4f9c-9fff-682aa5bba288.png) | ![4f9NA](https://user-images.githubusercontent.com/23270214/145108861-fcb41410-6e7d-4bcd-a569-73646ea363a2.png) |

greatmastermario:
Looks like there are some hardcoded references to max height (formerly 255) and/or max y for sections. Section data can now go as low as -4 and as high as 19 (inclusive), so any references to these boundaries will need to be looked for. I am taking a look, and if I find all the references I will open a PR.

greatmastermario:
https://github.com/overviewer/Minecraft-Overviewer/pull/2012

I got the fix working. Also fixes Issue 1954

Daanay:
@greatmastermario sorry to bother you. How I can use your bugfix/2003_1.18_Increased_Height with fixed render above 254 height to render my world?

greatmastermario:
@Daanay After you check out my branch, you need to recompile the C libraries. Follow the instructions in the documentation to do this. I did it with Python 3.9 with option 2 for compiling, but you should be able to use whatever works for you and the platform you are compiling on. Once that's done, run as normal.

Daanay:
Thank you so much!

jobe1986:
This bug does make me wonder about custom dimension types, which if I read correctly can have Y coordinates ranging from -2032 to 2031 (looking at min_y and height from custom dimension types on https://minecraft.fandom.com/wiki/Custom_dimension?so=search#Dimension_type ) Though as yet I havent tested how accurate that info is on the available options.

greatmastermario:
Since this issue was fixed with my PR and merged to the main branch, can we close this issue?

Issue #1954, opened 2021-06-28
Blocks below Y:0 are not rendering on the map generation


EpicnessTwo:
When generating using the latest build from GIT and the 1.17 Java file, blocks below Y:0 are not currently being rendered onto the map.

Example:
![image](https://user-images.githubusercontent.com/7784589/123709168-8420bf80-d864-11eb-81d8-8e0d8c98a8eb.png)

jobe1986:
Can confirm this is still the case as of the latest git commit.

EpicnessTwo:
@jobe1986 Sorry for the long delay, been so busy with other things. Still an issue on Minecraft Overviewer version 0.17.47 (7171af5)!

https://map.epiccraft.network/#/-25011/64/25037/-1/1_18_overworld%20-%20overworld/1_18_overworld

![image](https://user-images.githubusercontent.com/7784589/150364154-06135a6c-4ccd-44b0-9211-e71dbc683ce0.png)

morsik:
@EpicnessTwo of course it's still issue - this isssue was never closed or resolved so don't expect anything else.

As always in open source: Merge Request welcome :)

You can always try other branches that some people here tried (check for current opened Pull Requests) and verify them.

PR #2012, opened 2021-12-15
Bugfix/#2003 #1954 1.18 increased height

greatmastermario:
Fixed issues where any chunk sections below y=0 and above y=255 would not render properly after being added in 1.18

Issues:
https://github.com/overviewer/Minecraft-Overviewer/issues/1954
https://github.com/overviewer/Minecraft-Overviewer/issues/2003

Screenshots of working render attached:

Red wool is from y 0-192, Green wool is from 193-255, and Blue wool is from 256-319
image

All sandstone is generated below y=0 (sections Y=0 to -4)
image

### DIFF
--- a/overviewer_core/world.py
+++ b/overviewer_core/world.py
@@ -200,8 +200,8 @@ class World(object):
         ## clamp spawnY to a sane value, in-chunk value
         if spawnY < 0:
             spawnY = 0
-        if spawnY > 255:
-            spawnY = 255
+        if spawnY > 319:
+            spawnY = 319
             
         ## The chunk that holds the spawn location
         chunkX = spawnX//16
@@ -236,7 +236,7 @@ class World(object):
                 spawnY += 1
             # Next section, start at local 0
             inChunkY = 0
-        return spawnX, 256, spawnZ
+        return spawnX, 320, spawnZ
 
 class RegionSet(object):
     """This object is the gateway to a particular Minecraft dimension within a
--- a/overviewer_core/world.py
+++ b/overviewer_core/world.py
@@ -198,8 +198,8 @@ class World(object):
         disp_spawnZ = spawnZ = data['SpawnZ']
 
         ## clamp spawnY to a sane value, in-chunk value
-        if spawnY < 0:
-            spawnY = 0
+        if spawnY < -63:
+            spawnY = -63
         if spawnY > 319:
             spawnY = 319
             
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1101,10 +1101,10 @@ Depth
     **Options**
 
     min
-        lowest level of blocks to render. Default: 0
+        lowest level of blocks to render. Default: -64
 
     max
-        highest level of blocks to render. Default: 255
+        highest level of blocks to render. Default: 319
 
 Exposed
     Only renders blocks that are exposed (adjacent to a transparent block).
--- a/docs/design/designdoc.rst
+++ b/docs/design/designdoc.rst
@@ -29,7 +29,7 @@ tiles that can be displayed with a Leaflet interface. This section goes over how
 Minecraft worlds work and are stored.
 
 A Minecraft world extends indefinitely along the two horizontal axes, and are
-exactly 256 units high. Minecraft worlds are made of voxels (volumetric pixels),
+exactly 384 units high. Minecraft worlds are made of voxels (volumetric pixels),
 hereby called "blocks", where each block in the world's grid has a type that
 determines what it is (grass, stone, ...).  This makes worlds relatively
 uncomplicated to render, the Overviewer simply determines what blocks to draw
@@ -40,15 +40,15 @@ iteratively.
 The coordinate system for Minecraft has three axes. The X and Z axes are the
 horizontal axes. They extend indefinitely towards both positive and negative
 infinity. (There are practical limits, but no theoretical limits). The Y axis
-extends from 0 to 255, which corresponds with the world height limit. Each
+extends from -64 to 319, which corresponds with the world height limit. Each
 block in Minecraft has a coordinate address, e.g. the block at 15,78,-35 refers
 to 15 along the X axis, -35 along the Z axis, and 78 units up from bedrock.
 
 The world is organized in a three-layer hierarchy. At the finest level are the
 blocks (voxels). A 16x16x16 array of blocks is called a *chunk section*. A
-vertical column of 16 chunk sections makes a *chunk*. A chunk is therefore a 16
+vertical column of 24 chunk sections makes a *chunk*. A chunk is therefore a 16
 by 16 area of the world that extends from bedrock to sky. In other words, a 16
-by 256 by 16 "chunk" of the world. A 32 by 32 area of chunks is called a
+by 384 by 16 "chunk" of the world. A 32 by 32 area of chunks is called a
 *region*. Regions are stored on disk one per file.
 
 While blocks have a global coordinate (the ones you see in the debug output
@@ -355,7 +355,8 @@ origin being at the left corner.
 
 To ensure that block closer to the viewer are drawn on top while blocks that
 should be obstructed are drawn are hidden, the blocks are drawn one layer at a
-time from bottom to top (Y=0 to Y=15) and from back to front.
+time from bottom to top (Y=0 to Y=23, corresponding to sections Y=-4 to Y=19)
+and from back to front.
 
 From the data file on disk, block information in a chunk is a three-dimensional
 array of bytes, each representing a `block id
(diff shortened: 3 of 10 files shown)


greatmastermario:
Also fixes https://github.com/overviewer/Minecraft-Overviewer/issues/2011

overviewer_core/world.py
```diff
@@ -200,8 +200,8 @@ def find_true_spawn(self):
         ## clamp spawnY to a sane value, in-chunk value
         if spawnY < 0:
             spawnY = 0
-        if spawnY > 255:
-            spawnY = 255
+        if spawnY > 319:
+            spawnY = 319
```
ion1, line 204: Maybe the `spawnY < 0` check should also be updated to `-64`?

greatmastermario: Excellent point. I forgot to take that into account. Let me fix that.

CounterPillow:
Thanks for your work on this!

There is however one problem: if I take a 1.17 map, render it before this change, then play on it and render it with this change, I get all blocks shifted down some positions:

image

I'm not sure if we can reasonably do anything about this.

greatmastermario:
@CounterPillow Thank you for checking that. Is there a way for you to share the world download so I can download and test? I have a similar situation with one of my worlds, though in my case it is missing chunks. The 1.17 chunks that are rendering are properly aligned though.

image

greatmastermario:
Also, just to cover our bases, did you recompile the C libraries? I would assume you did, but I want to be sure in case it's just an issue of a missed step.

CounterPillow:
Yes, I did recompile the C libraries, Overviewer detects the changed version and wouldn't even have let me run it otherwise.

Here's the test world:
[117signs.zip](https://github.com/overviewer/Minecraft-Overviewer/files/7729078/117signs.zip)

It's important to render it first in overviewer git master, then launch the world, close it again so it writes some changes to the chunks as minecraft always does, then update the rendered map with your changes on top.

greatmastermario:
Thank you. I didn't think of the fact that I changed the version.

So that I'm clear on the instructions:

1. Render the world (which is from 1.17) in overviewer master branch
2. Open the world in Minecraft 1.18 and then close
3. Render the world in overviewer bugfix branch

If those are the instructions, I have a feeling we will need to require --forceupdate or --check-tiles to ensure the tiles are updated. The logic is not backward compatible, so forcing an update should resolve this problem, and we will need to notify users of this fact.

CounterPillow:
Step 2 is wrong, open the world in 1.17 so that it stays a 1.17 world. Other than that your test method is correct.

When I get the time I'll look into whether we can add a mechanism to forcerender if a map older than a certain version is detected.

ion1:
I wonder if it would be feasible to map positions into tiles such that the blocks in 1.17 worlds and upgraded 1.18 worlds align correctly?

greatmastermario:
I went ahead and added a version check to ensure there is a re-render on version update. It will still honor explicit render options, but it will default to forcerender if the version is changed or not present. I tested in local and it's working fine.

CounterPillow:
Hmmm, I'm not a huge fan of this.

1. *every* overviewer update will prompt rerenders, which is a problem for huge worlds
2. findGitVersion might not work on source tarballs
3. this probably still means that the coordinates are wrong

I think what ion said is the proper fix here; rendered block 0,64,0 should remain at 0,64,0 even with the new height changes.

This may be a matter of just a simple offset somewhere in the chunks-to-tiles code that was forgotten.

greatmastermario:
Got it. I will revert the last commit and continue to play with the offsets.

BafDyce:
If wanted, I can also check the patch with my huge map! My world is close to 1GB in size (`region` folder), roughly `6k x 6k` blocks and it contains chunks from various minecraft versions (oldest ones are from ~2011/2012). However, sharing world data (in case of incorrect rendering) will probably be impossible.

greatmastermario:
@CounterPillow I am still working through the code, but the offset solution is proving to be... problematic at best. Unfortunately, it is not as simple as subtracting 4 from the row as I thought it would be. There are four interconnected methods, and the math simply isn't enough to take care of the issue.

I can do one of two things:

1. I can continue to work through this and see if I can resolve the missing chunks section issue
2. I can modify the --forcerender version logic to address your concerns

Please note I also was unable to replicate the same issue on my end. There was no offset issue when I followed the steps you gave (render, load in 1.17, update render).

image

image

greatmastermario:
I went with option 2 for you to review. Adding the offset is proving to be very complicated. Here are the ways the new commit addresses the issue:

1. _every_ overviewer update will prompt rerenders, which is a problem for huge worlds: I added an explicit TILESET_VERSION which will only be changed if there is a breaking change, similar to how overviewer.h does it
2. findGitVersion might not work on source tarballs: Adding the explicit version to the code will make sure the version isn't corrupted or lost
3. this probably still means that the coordinates are wrong: You are right, I have fixed that in the JS file and tested to make sure it's working now.

Please let me know if you have any other concerns.

henu (approved):
I tested this fix to my world. You can see the result here: http://storage1.henu.fi/Henulandia3

The converted chunks, which previously didn't work, now render fine below y=0. Old chunks are also fine.

The chunks between converted and old are not rendered for some reason, but I understood it is out of scope of this PR. They were not rendered with master branch either.

greatmastermario:
@henu Thank you for the approval. I have found the issue with the missing chunks, but as you stated, I think that needs to be a separate PR/issue since it is a pre-existing issue in the master branch and not in scope for this PR.

Wolfensteijn:
@greatmastermario Have you checked Nether rendering with your code?

The height of the Nether (and also the End) has not changed, so depending on what exactly you changed, there might be issues there. In fact, I am seeing top bedrock on my Nether rendering, which it didn't have before applying this patch.

Map can be found here: https://map.spiritiacraft.org/#/2/64/-3/-6/SpiritiaCraft_Nether%20-%20nether/SpiritiaCraft%20-%20Nether

greatmastermario:
@Wolfensteijn I fixed it. Can you try again with the latest commit? You will need to rebuild the C libraries for the change to take effect.

Wolfensteijn:
@greatmastermario yes, that fixed it!
https://map.spiritiacraft.org/#/-116/64/-138/-2/SpiritiaCraft_Nether%20-%20nether/SpiritiaCraft%20-%20Nether

JakobDev:
The height limit can be increased with a data pack as [AntVenom showed in his video](https://www.youtube.com/watch?v=tH76tx-AnWI&t=79s), what means the height is not hardcoded in 1.18 as in the versions below. So I think it's not a good idea to use a hardcoded value in Overviewer. Players will maybe increase the height by data packs for larger buildings. And it should be possible to view the full building in Overviewer in that case.

greatmastermario:
@JakobDev There are a couple problems with having a variable setup in Overviewer that need to be weighed to support variable height limits, and I think they need to be in a separate issue discussion so the admins for this repo can make a decision:

1.  What is the scope for Overviewer worlds? Is it just vanilla and texture packs, or is it all possible mods and data packs? My inclination is that it is the former, and that since this is open-source we can modify the source for specific use cases outside the original scope. @CounterPillow Can you or someone else from the Overviewer org comment on this point?
2. Adding variable heights is technically feasible if we were just dealing with the overall height, but there are other things that need to be considered such as specific Y values relative to the tileset (such as spawn point) that would need to be determined by the data pack. Data pack support is currently not present in Overviewer, so this would need to be opened as an issue with a brand-new feature and development.

For now, my vote is for this fix to be included on its own, and we can talk about a variable world height feature in a separate issue and PR.

dalandis:
I ran your edits for my map and see that the nether world is broken.

example http://map.gmgame.ru/#/-48/64/263/-5/gmgame_nether%20-%20nether/nether

Works if you turn on the roof. Probably due to height adjustment - Depth(min=0, max=50)
Снимок экрана 2022-01-13 в 17 12 23
Снимок экрана 2022-01-13 в 17 12 31

dalandis:
yes, I'm talking about the depth options.

my config
`
nether_smooth_lighting = [Base(), Depth(min=0, max=50), EdgeLines(), Nether(), SmoothLighting(strength=0.5)]
renders["nether"] = {
     "world": "gmgame_nether",
     "title": "Nether",
     "rendermode": nether_smooth_lighting,
     "dimension": "nether",
     "crop": (-1562, -1562, 1562, 1562),
     'defaultzoom': 5
}
`

before the map looked like this

greatmastermario:
@dalandis Fixed. You can pull the latest commit and it should work now. I tested on my world, and it works.

dalandis:
I restarted rendering. But I can't confirm yet, we have to wait

dalandis:
It seems to work. Thank you. It takes a very long time to complete rendering.

Снимок экрана 2022-01-14 в 12 14 23

KosmicAnomaly:
On the topic of hard-coding a max height, there are datapacks such as [Amplified Nether](https://www.planetminecraft.com/data-pack/amplified-nether-1-18/) which increase the height of the Nether, which then breaks all Nether renders. I think that the Nether roof should be able to be specified to solve this issue.

Atten007 (approved):

wolfen351:
@CounterPillow Respectfully, could you please have a look at this and include it if it meets the quality standards?

Issue #2010, opened 2021-12-15
Multiple chunks fail to render in 1.18.1


drnixon:
Built overviewer from latest (7171af5) on macOS 12.1. Rendering into a clean directory after using --forceUpgrade on all chunks in a Spigot server. Getting a repeated error with multiple chunks:
```
ERROR:root:Could not render chunk -90,70 for some reason. This is likely a render primitive option error.                                                    ] 82678 106.43t/s eta 11h 24m 54s
ERROR:root:Full error was:
Traceback (most recent call last):
  File "/Users/[username]/Documents/MC/overviewer/Minecraft-Overviewer/overviewer_core/tileset.py", line 1091, in _render_rendertile
    c_overviewer.render_loop(
  File "/Users/[username]/Documents/MC/overviewer/Minecraft-Overviewer/overviewer_core/world.py", line 2120, in get_chunk
    retval = super(CachedRegionSet, self).get_chunk(x,z)
  File "/Users/[username]/Documents/MC/overviewer/Minecraft-Overviewer/overviewer_core/world.py", line 1935, in get_chunk
    return self._r.get_chunk(x,z)
  File "/Users/[username]/Documents/MC/overviewer/Minecraft-Overviewer/overviewer_core/world.py", line 1713, in get_chunk
    chunk_data = chunk_data['Level']
KeyError: 'Level'
```
Related to `chunky` now exceeding 255? End and nether seem to be rendering correctly, errors appear to be only in overworld renders so far.

sirl1on:
I am getting the same error. No export is generated at all

- Vanilla Server
- World generated in 1.16, now at 1.18.1
- Using latest Docker Container setup without custom config

`docker run --rm -v ~/.minecraft/saves:/tmp/server/:ro -v $(pwd)/export/:/tmp/export/:rw -it marctv/minecraft-overviewer`

```
2021-12-22 09:51:15  Welcome to Minecraft Overviewer version 0.17.39 (2402e41)!
2021-12-22 09:51:16  Generating textures...
2021-12-22 09:51:25 E An error has occurred. This may be a bug. Please let us know!
See http://docs.overviewer.org/en/latest/index.html#help

This is the error that occurred:
Traceback (most recent call last):
  File "/tmp/overviewer/overviewer.py", line 653, in <module>
    ret = main()
  File "/tmp/overviewer/overviewer.py", line 562, in main
[...]

PR #2040, opened 2022-05-27
Correcting Level error in upgraded 1.18 worlds

drnixon:
Edit suggested by Zevac-ZSF in #2010 corrects render errors that cause the bug.

### DIFF
--- a/overviewer_core/world.py
+++ b/overviewer_core/world.py
@@ -1710,7 +1710,8 @@ class RegionSet(object):
         if not 'sections' in chunk_data:
             # This world was generated pre 21w43a and thus most chunk data is contained
             # in the "Level" key
-            chunk_data = chunk_data['Level']
+            if 'Level' in chunk_data:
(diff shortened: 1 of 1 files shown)


overviewer_core/world.py
```diff
@@ -1710,7 +1710,8 @@ def get_chunk(self, x, z):
         if not 'sections' in chunk_data:
             # This world was generated pre 21w43a and thus most chunk data is contained
             # in the "Level" key
-            chunk_data = chunk_data['Level']
+            if 'Level' in chunk_data:
+            	chunk_data = chunk_data['Level']
```
ekeih, line 1714: This seems to miss a few spaces? :thinking: 
```suggestion
            	   chunk_data = chunk_data['Level']
```

CounterPillow:
Fixed this differently in 4e51b216524ccc535893efdaeef48d75061640d4 instead, the problem is that a compatibility fix is applied when it shouldn't be, not with the compatibility fix itself.

PR #2044, opened 2022-06-10
WIP: Begin supporting Minecraft Bedrock Edition maps

### LINKED ISSUE #921: request support for pocket and bedrock edition maps
$ overviewer.py --rendermodes=smooth-lighting "world1" ./
2013-04-24 16:29:56  Welcome to Minecraft Overviewer!
2013-04-24 16:29:56 E An error has occurred. This may be a bug. Please let us know!
See http://docs.overviewer.org/en/latest/index.html#help

This is the error that occurred:
Traceback (most recent call last):
  File "/usr/bin/overviewer.py", line 532, in <module>
    ret = main()
  File "/usr/bin/overviewer.py", line 413, in main
    w = world.World(render['world'])
  File "/usr/lib/python2.7/dist-packages/overviewer_core/world.py", line 97, in **init**
    data = nbt.load(os.path.join(self.worlddir, "level.dat"))[1]['Data']
  File "/usr/lib/python2.7/dist-packages/overviewer_core/nbt.py", line 29, in wrapper
    return func(fileobj, *args)
  File "/usr/lib/python2.7/dist-packages/overviewer_core/nbt.py", line 37, in load
    return NBTFileReader(fileobj).read_all()
  File "/usr/lib/python2.7/dist-packages/overviewer_core/nbt.py", line 181, in read_all
    tagtype = ord(self._file.read(1))
  File "/usr/lib/python2.7/gzip.py", line 256, in read
    self._read(readsize)
  File "/usr/lib/python2.7/gzip.py", line 291, in _read
    self._read_gzip_header()
  File "/usr/lib/python2.7/gzip.py", line 185, in _read_gzip_header
    raise IOError, 'Not a gzipped file'
IOError: Not a gzipped file

barrett777:
I'm working on adding support for Bedrock maps (Fixes #921). This pull request is not ready for merging, I'm just opening it as a spot to show my code and ask for help.

There's a couple areas that need more development, and I don't have as much time available to finish it.  Is anyone able to help?

### Code changes

Nearly all my changes are in `nbt.py` and `world.py`:

Bedrock data is stored in one leveldb database - there's no concept of RegionFiles, so I just store everything in one 'RegionFile' in Overviewer.

Most of my code changes are making Overviewer work correctly with only one RegionFile, and reading from the Bedrock leveldb database

It's worth noting I'm not planning to use Bedrock textures, so users still need to have Java edition installed.

Imported code
Mojang leveldb dlls via BSD3-Clause: https://github.com/Mojang/leveldb-mcpe
This library for reading chunks and blocks from leveldb: https://github.com/BluCodeGH/bedrock

Here's a sample test Bedrock map: https://drive.google.com/file/d/1JIxZEfdAgee4-_Irn8Zb5vYUMbtmkqD9/view?usp=sharing

To run Overviewer on a Bedrock map, you need to set config parameter 'isbedrock = 1', and specify the full directory path to a Bedrock map folder.  You also need to set 'processes = 1' (see below)

More info on Bedrock Edition leveldb key/value formats: https://minecraft.gamepedia.com/Bedrock_Edition_level_format

### Known issues I need help with, in order of importance

I can't check when a chunk was last modified. For now 'get_chunk_timestamp' just returns the current time. A different Bedrock mapping software instead calculates a hash for each chunk to detect changes. This is a non-trivial change because Overviewer currently uses rendered image file modified times. To instead use hashes, we would need to probably save a json file in the render directory with a dictionary of all chunk coordinates, and a hash of each chunk. Then Overviewer would need to read from that to determine what chunks have been changed. We would also need to add a hashing function to 'Bedrock.py' for the Chunk class, and make sure it also hashes relevant chunk data such as subchunk, blocks, biome data, etc.

My included implementation of leveldb database access doesn't support multiple processes, so at this time you need to set 'processes' to 1 or it will blow up.  I *think* it supports multiple threads, but I'm skeptical.  This definitely makes performance pretty terrible.  I haven't yet fully rendered a big map.  I have a 700mb map that estimated to take ~300 hours to generate.  There is probably a clever solution for managing database access across processes, or even just making 10 copies of the map save file and using one process for each.

I need help "converting" Bedrock's block types to Anvil, so blocks render the correct texture. I can find each block's text type such as 'minecraft:bedrock', and Anvil and Bedrock share most of these (hooray!). There are still a lot of unrecognized block types though. For now I'm replacing any unrecognized blocks with 'minecraft:bedrock'. I could use help mapping all the block types that don't match. Maybe this could be copied from any of the existing map converting software, for example https://github.com/gentlegiantJGC/PyMCTranslate (See world.py line 1104)

I can't yet find the 'true spawn', so I just return (0, 0, 0). I think this is stored in 'level.dat'. However this is a slightly different format than Anvil (See world.py line 180)

Bedrock does store chunk 'biome' data, but I don't know what that is. At this time I'm just passing the Bedrock chunk biome data through the existing Overviewer biome code, and it doesn't blow up, so maybe that is okay.  It would be nice if someone who understands the chunk biome bytes could take a look.

If you add multiple runs across different dimensions, you get some crossover between them - some Overworld chunks get rendered in the Nether render, etc. I'm guessing it's a caching issue somewhere due to the fact that Bedrock databases only have one regionfile. For now I just enforce that you can only render one dimension in a config file (you can have multiple renders in the same dimension, and it works fine).

The map save file database can get corrupted if you close Overviewer while it is running.  This probably needs to be fixed, but I'm not sure how.  For now, make sure you only run on a backup of your map.

### Known issues that probably can't be helped

I'm not sure if Bedrock stores skylight or blocklight data. For now I'm just defaulting skylight to full.

I think Minecraft Bedrock Edition doesn't save chunks to disk unless something is modified, so there can be blank gaps on the map for explored areas where a player just runs through

I think the map save file database can get corrupted if you play Minecraft in this world while Overviewer is running.  I don't think this can be fixed.


patricksmalley:
Would it be possible for the script running Overviewer to copy the Minecraft world, then analyze the chunks? Once it is done the temporary copy can be deleted. This may fix the issue that occurs when analyzing a live world. 

This is what I am doing for Bedrock Viz on a Bedrock server and it is working well. 

I miss Overviewer and would love to have this running on my Bedrock server!

JoshuaMarti:
You should join the Discord and bring this up there! https://discord.gg/32Bz2yW

This repo has been abandoned, but the project is being continued at https://github.com/GregoryAM-SP/The-Minecraft-Overviewer/

Issue #1538, opened 2019-03-16
SIMD acceleration


Wunkolo:
A lot of the [core critical C code](https://github.com/overviewer/Minecraft-Overviewer/tree/master/overviewer_core/src) in relation to image and block manipulation is implemented in a very serial matter though processors these days allow for [vectorized instruction sets](https://en.wikipedia.org/wiki/SIMD) where multiple fragments of data can be processed in parallel rather than one at a time. The x64 instruction set architecture includes up to SSE2 and I feel that a lot of the C code can take advantage of these instruction sets though [SIMD intrinsics](https://software.intel.com/sites/landingpage/IntrinsicsGuide/) exposed through headers like `immintrin.h` and such to possibly get huge improvements in speed for certain operations like image blitting and block/tile iteration and bit/byte manipulation

I might look to contribute once I get a good feel of the code base but I am wondering how you would feel about having PRs that use these intrinsics to vectorize and speed up critical sections of code as this would make some assumptions about the minimum requirements for the end-user. In particular it will require that the user be on a 64 bit computer which is kind of reasonable in 2019 or it can be something determined at compile time so that these intrinsics are not used unless the user opts in for them(Which in that case, will allow the user to opt-in for even more vectorization extensions such as AVX2 or AVX512).

CounterPillow:
Determine it at compile time, juggling around function pointers of hot functions at runtime based on cpuid is just going to prevent inlining, which is bad. Personally I'd either target SSSE3 or AVX2. We still support 32-bit x86 for the time being, and we also support non-x86 based architectures such as ARM or AArch64.

I won't merge anything that isn't properly profiled and benchmarked though to ensure that it actually does make things faster.

Wunkolo:
Sounds fair, gcc and clang both provide compile-time preprocessor definitions for the target architecture, though windows only defines them up to AVX2 I think
```
gcc -march=skylake-avx512 -dM -E - < /dev/null | egrep "SSE|AVX"
#define __AVX512F__ 1
#define __SSE4_1__ 1
#define __SSE4_2__ 1
#define __AVX512BW__ 1
#define __SSE2_MATH__ 1
#define __AVX__ 1
#define __AVX512VL__ 1
#define __AVX512CD__ 1
#define __AVX2__ 1
#define __SSE_MATH__ 1
#define __AVX512DQ__ 1
#define __SSE2__ 1
#define __SSSE3__ 1
#define __SSE__ 1
[...]

PR #2045, opened 2022-06-15
block_class: Complete x86-SIMD

Wunkolo:
Accelerates `block_class_is_subset` with all available x86 simd features
up to AVX512.

This is to complete the scope of #1538


CounterPillow:
Tested with `CFLAGS="-march=native -mtune=native -mavx2" ./setup.py build` on a Zen+ laptop and it provided like a 6% speedup on a fairly small map so I'd say this is good and nice on any CPU that actually does AVX2 properly. Thanks!

Issue #1899, opened 2021-02-01
Black surface of the render 


Dellism:
I'm trying to make a render using data from 1.16.4 server using world-downloader. Render modes normal and cave works perfect, but smooth-lightning and lightning makes surface completery black excluding few blocks. I used this command: ./overviewer.exe "world folder" "render folder" --rendermodes=smooth-lighting.
![image](https://user-images.githubusercontent.com/61477283/106515518-7ac23f80-64de-11eb-97ae-1ed0ddeaf9c1.png)

gP387:
I have the same issue with smooth_lighting and with smooth_night it doesn´t render any light sources.
It is a 1.15.2 Map downladed from a server with World Downloader
this is an other rendering that I made before(it wasn´t completly finished, the missing chunks are my fault)
![image](https://user-images.githubusercontent.com/75202782/108852073-6f49cc00-75e5-11eb-816b-3a8fb0527c9b.png)

with smooth_lighting
![image](https://user-images.githubusercontent.com/75202782/108851409-aec3e880-75e4-11eb-915e-16134313e4d6.png)

with smooth_night (very dark)
![image](https://user-images.githubusercontent.com/75202782/108851531-d1560180-75e4-11eb-943f-17c77b0a9392.png)

32Vache:
same issue for me, a solution i found is to open the map in singleplayer, max out render distance and fly around, all chunks loaded will get thier lightning fixed on the render, but it can be very trippy to do on large maps, so would be good if we could get this bug fixed.

Issue #2035, opened 2022-04-26
Some chunks are shown as not explored/not rendered


ubuntuchosen:
Hello, didn't find this issue here, in case it's a duplicate, please comment the # of the original issue.

I am occasionally running overviewer to generate my local server's map, somewhere in 2021 I last generated the map, with all parts of the map showing correctly (this was 1.16 or 1.17). Since updating to 1.18.2, some chunks are rendered as unexplored/not rendered. 

You can compare map from 2021 to map from 2022:
2021 - https://t0day.eu/minecraft/210710-normal/index.html
2022 - https://t0day.eu/minecraft/220418-normal/index.html

I checked most parts ingame, explored them with a friend and we see all blocks there are intact, as they should be. However, even the next render I tried, I still see the chunks on parts of the map as black, un-rendered parts.

Anyone know what's going on or is also experiencing the same issue?

ralokt:
This is probably #2002 - I'm making a PR that addresses this right now!

Issue #2029, opened 2022-02-25
Lighting issue


LavaCaster:
I converted the world from bedrock to java and then I did the render so thats probably the issue.
Only somme chunks have lighting and other are in complete darkness. Heres how it looks https://prnt.sc/7ha6VRUAKcwt
sorry for bad inglish

alecrux:
It looks like it's a duplicate of #1899
I'm having the exact same issue, except my world has been upgraded from 1.18.2 to 1.19.

The light is properly rendered only in chunks that I visited in 1.19.
![image](https://user-images.githubusercontent.com/35157010/178089528-932a58cf-e364-4d85-9014-e0b104f476fd.png)


This is something I observed when testing the upgrade of my server, which consists in updating the jar file and running the server with the --ForceUpgrade flag the first time.
I'm happy to run any additional test you might think of as I won't update my production server until that issue is resolved or I find a workaround that does not consist in visiting the chunks in-game.

alecrux:
It looks like it's a duplicate of #1899
I'm having the exact same issue, except my world has been upgraded from 1.18.2 to 1.19.

The light is properly rendered only in chunks that I visited in 1.19.
![image](https://user-images.githubusercontent.com/35157010/178089528-932a58cf-e364-4d85-9014-e0b104f476fd.png)


This is something I observed when testing the upgrade of my server, which consists in updating the jar file and running the server with the --ForceUpgrade flag the first time.
I'm happy to run any additional test you might think of as I won't update my production server until that issue is resolved or I find a workaround that does not consist in visiting the chunks in-game.

alecrux:
A good workaround consists in upgrading the server **without** the --ForceUpgrade argument.
I have not found any solution to recover from a server that has been updated with the --ForceUpgrade argument yet.

Issue #2024, opened 2022-01-22
Overviewer drawn only small part of chunks


Staric1982:
I have a small world with long "ways out". Overviewer rendred them, but after converting to 1.18 (i think so), it cant draw more than small part around a base. Screenshots attached.
Map directory was cleared, but no success.

Overviewer result:
<img width="857" alt="Screenshot 2022-01-22 at 18 56 54" src="https://user-images.githubusercontent.com/3616310/150646221-5dae53b9-c5e5-4ec1-b776-830b9dd10149.png">

Another rendered result (true map)
<img width="934" alt="Screenshot 2022-01-22 at 18 57 05" src="https://user-images.githubusercontent.com/3616310/150646230-18392b9b-e18b-45e1-9cbf-c95f69181080.png">

Overviewer config:
<img width="994" alt="Screenshot 2022-01-22 at 18 58 30" src="https://user-images.githubusercontent.com/3616310/150646233-60158579-1899-4f3c-bde6-66652f527016.png">

dalandis:
looks like it's a duplicate https://github.com/overviewer/Minecraft-Overviewer/issues/2002

Staric1982:
Looks like the same, but a little. Ok, wait for update. Thanks.

PR #2060, opened 2022-07-28
Fix dark chunks - addresses #2029 and #1899, prepares for fix of #2002, #2024, and #2035

ralokt:
### The Problem

It is possible for chunks to have no sky light data whatsoever.

This can happen for a number of reasons (see also: #2029, #1899), but in particular, all worlds that have been optimized with `--forceUpgrade` in 1.19 will be completely dark until chunks are loaded ingame.

Also, Minecraft sometimes saves chunks with incomplete light data. These chunks only contain "spillover" light from neighboring chunks, but the sky light from the chunk itself is not taken into account. It is common for these chunks to appear on "version borders", and they are almost completely dark. Screenshots of an example at two different zoom levels:
Screenshot_20220729_005120
Screenshot_20220729_005036

### Overview of contents

This PR addresses this issue in two steps:
 - `d521770640198be64061dc2a610dbcdc914477ef` checks all chunks for a section with sky light data. If no such section exists, the chunk will be fully brightened.
 - `57727aead9d612fda22a1c32b5cecd55745e669d` additionally checks chunks that do have sky light for a block with full sky light. If no such block exists, it will assume that the chunk is mostly dark and fully brighten it (detailed explanation below).
 
Also, in order to cleanly implement this change, I did a bit of refactoring.

### Detailed reasoning for the decision to check chunks for a block with full skylight

Two observations:
 - A chunk without "own" skylight that only has skylight data from neighboring chunks "pour in" will never have skylight 15 anywhere.
 - Most chunks with skylight will have full skylight at the top.

These observations together mean that:
 - The heuristic we chose is very good - there are edge cases like a leaf roof at the build limit that would break, but I'd argue that these edge cases are less important than the ugly borders pictured above
 - It is also very fast if we go from top to bottom.
 
In practice, on my test world of slightly over 5000 tiles, a pure python PoC implementation of this heuristic bumped the runtime from 8:35 to 8:50, and the numpy implementation somehow managed to finish rendering faster, in 8:06. While that is based on only one run per implementation, I think it's safe to say that this change does not significantly affect performance.

And in any case, if it is deemed preferable not to do that, we can just yeet the second commit.

### Reverse dependency to rendering 1.18+ chunks properly again

This issue came up in #2051, my first attempt at fixing the rendering of pre-1.18 worlds that were loaded in Minecraft 1.18 or higher. Since these chunks turn out completely dark if they were optimized but not loaded in 1.19 (see above), it makes sense to first fix this lighting issue before making these chunks render.

@cliffmeyers and I agreed to split the change into two PRs to avoid scope creep, so I'd like to get this merged first.

### DIFF
--- a/overviewer_core/world.py
+++ b/overviewer_core/world.py
@@ -1719,10 +1719,23 @@ class RegionSet(object):
         fullbright = chunk_data.get("Status", "") == "spawn" and 'Lights' in chunk_data
         if not fullbright and self.get_type() is None:
             # In the overworld, we also default to fullbright if these is no sky light
-            # in the chunk whatsoever
-            for section in sections:
+            # in the chunk at full level.
+            # We can't just check for the complete absence of light, as Minecraft will
+            # sometimes partially generate sky light in chunks that is just spillover
+            # from neighboring chunks, but doesn't take the chunk itself into account,
+            # leaving us with a mostly dark chunk.
+            # The nice thing about sky light that spills over is, it is never at full
+            # level.
+            # Also, almost all chunks will have full sky light at the very top.
+            # Therefore: if a chunk contains sky light, but none of it is full, we can
+            # be 99% sure that that chunk is one of these mostly dark ones.
+            # If you've built a roof of leaves at build limit and are reading this:
+            # I'm sorry.
+            for section in sections[::-1]:
                 if "SkyLight" in section:
-                    break
+                    light = numpy.frombuffer(section["SkyLight"], dtype=numpy.uint8)
+                    if numpy.any((light>>4) == 15) or numpy.any((light & 15) == 15):
+                        break
             else:
                 fullbright = True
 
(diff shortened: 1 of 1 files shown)


cbs228 (approved):
I have tested this on a large 1.19.3 world that has been around since before 1.16. An `overviewer.py --forcerender` completed without issues and fixed #1899 for me. As expected, chunks without lighting info render at full brightness. This is vastly preferable to full darkness.

I did a brief code review and didn't find anything major, but someone with more experience with this codebase than me should probably look at it.

It is my understanding that Minecraft-Overviewer is no longer maintained, so it is unlikely that this PR will be merged. I recommend that any future maintainers and/or forks should merge this.

cliffmeyers:
@cbs228 thanks for sharing this. I'm going to pull this branch and test it against my ~8 GB when I get a chance. I would love to get a decent full render of my 1.19 world before bumping to 1.20.

Issue #2042, opened 2022-05-30
Higher zoom level than 8?


dkim19375:
How do I increase the zoom level? (Since the highest I can get is zoom level 8 which gives a 12x12 block resolution)
Changing `overviewerConfig.js` just makes higher zoom levels black.
Changing `config.py` and adding the line `maxzoom = 9` doesn't seem to change anything either.

In the docs, it says:
> Note: This does not change the number of zoom levels rendered

but then how do I change the number of zoom levels rendered? Or is zoom level 8 the maximum?

CounterPillow:
The resolution at which the most zoomed in level is rendered is fixed, there is no way to change it and it wouldn't be easy to implement this due to how overviewer is designed.

PR #2061, opened 2022-08-04
feat: one extra virtual zoom for close up inspections

granis:
I often find myself wanting more zoom, and in there is support for this in Leaflet.
https://leafletjs.com/reference.html#gridlayer-maxnativezoom
Also, ping #2042.

For example map see https://maps.xtralarge.se/#/-10535/64/10305/-3/Terranova/terranova_day


BenKennish:
i'd love this feature

matrixhacker:
You added maxNativeZoom but it doesn't appear to be used anywhere. What is it's purpose?

PR #2091, opened 2023-05-15
Fix ValueError: x1 must be greater than or equal to x0

conqp:


adamenger:
this did the trick. 

👏 to @conqp

PR #2095, opened 2023-07-04
Add textures from 1.20

MinerPL:


MinerPL:
I love github and they fucked UI

PR #2112, opened 2025-10-08
Issue2013 dist utils removal

greatmastermario:
PR to remove distutils, which is not available in recent Python versions.

Opening as a draft due to issues with rendering. I am working on those and will mark the PR ready for review after those are fixed. Leaving this here for comment and possible assistance fixing the issues.

Currently the render runs, but no textures are generated in the output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ settings*.py
 # header files that may be copied over, if missing
 ImPlatform.h
 Imaging.h
+Arrow.h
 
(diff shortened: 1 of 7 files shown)


JoshuaMarti:
Hello! Just so you know, this project has been abandoned by the original maintainers, but we're continuing it at https://github.com/GregoryAM-SP/The-Minecraft-Overviewer

